PR #23, opened 2023-02-13
Delete github.svg:Zone.Identifier

Czompi:
I removed `github.svg:Zone.Identifier` file, because I can't clone the repo on Windows OS (Windows use `:` for drive letters).
From my quick research, it just provides metadata for the image, to be able to tell where it originates (viewing file in github online file viewer confirms that).

### DIFF
--- a/client/public/images/icons/github.svg:Zone.Identifier
+++ b/client/public/images/icons/github.svg:Zone.Identifier
@@ -1,3 +1,0 @@
-[ZoneTransfer]
-ZoneId=3
-HostUrl=https://fontawesome.com/


martinkadlec0:
Oh, thanks. I use WSL to develop stuff and whenever I copy any image files it automatically creates these stupid identifiers. Sometimes I forget to remove them.

Czompi:
I just found an another one: `Grass_Block_JE1.png:Zone.Identifier`
It seems that these 2 were the only ones, that you missed.
Sorry about that.

PR #26, opened 2023-02-14
Added LocalStorage to save and load decks

ijzm:
Allows the user to Save and Load your Loadout.
To save:
1. Create your deck
2. Add a name to it
3. Click the "Save" Button

To load:
1. Write the name of the deck
2. Click "Load"

TODO (Maybe in another PR): Save your current deck to a custom LocalStorage. At the begining of the game, check if that key exists. If it exists, load it automatically

### DIFF
--- a/client/src/app/deck/deck.tsx
+++ b/client/src/app/deck/deck.tsx
@@ -49,6 +49,8 @@ const Deck = ({setMenuSection}: Props) => {
 		}))
 	)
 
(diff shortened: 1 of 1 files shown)


martinkadlec0:
Thank you, I think eventually I'd like to replace the input with something nicer, but for now this is amazing so people don't have to make their decks all the time :)

PR #34, opened 2023-02-15
Updated questions readme about mending effect

keckrich:


martinkadlec0:
Hey the question was about where in the deck the card should go. It was always clear it should not be your hand.

PR #42, opened 2023-02-17
add Dropdown for deck selection

Xarlus:
**Purpose**
The dropdown should enable the user to select the deck without remembering all strings.

**additional infos**
what is not working is populating the list becaus of noide.remove() is not working in TS.
everything else should be working fine


martinkadlec0 (changes requested):
I can't approve this as it is sorry. It needs to be udpated to use react mechanics.

martinkadlec0:
We have now a datalist for similar puprose.

PR #51, opened 2023-02-18
Fix gems ability to allow placing regular effect card after attack

Sense101:


martinkadlec0 (changes requested):
Discussed on discord. Gem is bit more complex to fix.

martinkadlec0:
Closing for now, since the implementation needs to be reworked.

PR #52, opened 2023-02-18
Save default deck between sessions

ijzm:
Whenever you finish making your deck, save it as "Loadout_default"
On main menu, check if "Loadout_default" exists, if it does, update your deck to use it.


martinkadlec0 (changes requested):
This will rewrite custom decks with default every time user return to menu. Also it makes server generate custom deck unnecessarily. (more in Discord)

Sense101:
This has now been implemented properly in the latest update. Thanks for the PR though!

PR #55, opened 2023-02-19
very rough cut at a personal, oauthed, win-loss counter

AndyNovo:


client/src/logic/fbdb/fbdb-reducer.ts
```diff
@@ -0,0 +1,27 @@
+import {AnyAction} from 'redux'
+
+type FbState = {
+	uuid: string
```
martinkadlec0, line 4: Is there a reason we now ahve the uuid/stats state in both session & fbdb reducer?

client/src/app/app.tsx
```diff
@@ -4,6 +4,7 @@ import {getPlayerName} from 'logic/session/session-selectors'
 import {getGameState} from 'logic/game/game-selectors'
 import {getStatus} from 'logic/matchmaking/matchmaking-selectors'
 import LostConnection from 'components/lost-connection'
+import {authlogin, statsupdate} from 'logic/session/session-actions'
```
martinkadlec0, line 7: Super nitpick, but lets use camelcase here O:)
statsupdate -> statsupdate
authlogin -> auth:ogin

client/src/app/app.tsx
```diff
@@ -20,6 +21,23 @@ function App() {
 	const socketStatus = useSelector(getSocketStatus)
 	const [menuSection, setMenuSection] = useState<string>('mainmenu')
 
+	firebase.auth().onAuthStateChanged(user => {
```
martinkadlec0, line 24: Can you please rebase on latest master & run pretiter on your code? There is lots of formatting stuff that prettier will fix. On latest master there is a pre-commit hook that will run prettier for you. Alternatively `npx prettier --write server/. client/src/.` should work.

server/logs/index.js
```diff
@@ -1,35 +1,82 @@
+import { createRequire } from 'module';
+const require = createRequire(import.meta.url);
+var serviceAccount = require("../adminKey.json");
```
martinkadlec0, line 3: Reaplce all `var` with `const` (or `let` if its not constant).

server/logs/index.js
```diff
@@ -1,35 +1,82 @@
...
+        //"gameoverstat";
+        let numGames = Object.values(server.games).filter(gameRecord => !!gameRecord.task).length;
+        this.gamecountref.set(numGames-1);
 			})
 		})
 
 		server.hooks.playerJoined.tap(this.id, (playerInfo) => {
+			let numPlayers = Object.keys(server.allPlayers).length;
+      this.playercountref.set(numPlayers);
 			console.log('player joined: ', playerInfo.playerName)
```
martinkadlec0, line 73: Let's get rid off the example console.log so they dont flood the console.

martinkadlec0, line 73: Let's get rid off the example console.log so they don't flood the console.

client/src/logic/socket/socket-saga.ts
```diff
@@ -52,6 +52,24 @@ function* socketSaga(): SagaIterator {
 		socket.on('connect', connectListener)
 		socket.on('disconnect', disconnectListener)
 		socket.on('connect_error', connectErrorListener)
+		socket.on("gameoverstat", (result)=>{
```
martinkadlec0, line 55: Can we move this to `logic/fbdb/fbdb-saga` ? You can import `receiveMsg` from scoket-saga and then wait for new message like:
```javascript
function* fbdbSaga() {
  while (true) {
    const newMessage = yield call(receiveMsg, 'gameoverstat')
    // update stats
  }
}
```
You can fork it in root-saga after the socket saga:
```javascript
yield fork(socketSaga)
yield fork(fbdbSaga)
```

martinkadlec0 (changes requested):

PR #56, opened 2023-02-19
Card design changes

Razboy20:
This PR mostly focuses on bringing the current design of the card to be more inline with how the originals are designed.

Changes:
- Replace star icon with a version traced from Hermitcraft 
- Add white icon outlines (imo better than nothing, but the edges are currently slightly rounded, which cannot be fixed in-browser, and necessitates modification of the icon images.
- Increase card text kerning & Improve google font preloads


client/src/components/card/hermit-card-svg.module.css
```diff
@@ -3,7 +3,7 @@
 
 .card {
 	background: white;
-	letter-spacing: 1px;
+	letter-spacing: 0.2rem;
```
martinkadlec0, line 6: Please avoid rem units, they aren't used anywhere else so might cause unexpected behavior when someone plays with font-sizes.

martinkadlec0 (changes requested):

client/src/components/card/hermit-card-svg.module.css
Razboy20: Ah sorry; I meant to use em rather than rem, but em is the proper unit of size for text adjustments.

Razboy20:
If it's somehow possible to achieve a nicer looking white border on the card icons' in-browser, that would probably be best, however I can't think of any good method to get such a result. If the borders need to be improved, then the source images will have to be manually edited.

martinkadlec0 (approved):

martinkadlec0:
Accepted, though If I really hope I will never need to understand that svg filter ...

PR #57, opened 2023-02-19
Reorder validation and enforce at least one hermit in the deck

Sense101:


server/utils/index.js
```diff
@@ -147,11 +158,7 @@ export const validateDeck = (deckCards) => {
 	})
 
 	if (tooManyDuplicates)
-		return 'No card other than items cards can be more than 3 times in your deck.'
+		return 'You cannot have more than 3 duplicate cards unless they are effect cards.'
```
martinkadlec0, line 161: This should still be item cards I think.

martinkadlec0 (changes requested):
Just that one thing, otherwise looking good.

martinkadlec0 (changes requested):
Just that one thing, otherwise looking good.

martinkadlec0 (changes requested):
Just that one thing, otherwise looking good.

martinkadlec0 (approved):
Just that one thing, otherwise looking good.

PR #59, opened 2023-02-19
Anonstats

AndyNovo:
Reduced the scope, created fbdb folders, cleaned the stats


client/src/logic/session/session-reducer.ts
```diff
@@ -23,6 +23,10 @@ const loginReducer = (
 	switch (action.type) {
 		case 'LOGIN':
 			return {...state, connecting: true}
+		case 'STATS':
```
martinkadlec0, line 26: This should be removed I think right?

client/src/root-saga.ts
```diff
@@ -3,12 +3,14 @@ import {SagaIterator} from 'redux-saga'
...
 import {loginSaga, logoutSaga, newDeckSaga} from 'logic/session/session-saga'
 import matchmakingSaga from 'logic/matchmaking/matchmaking-saga'
+import fbdbSaga from 'logic/fbdb/fbdb-saga'
 
 function* appSaga(): SagaIterator {
 	yield call(loginSaga)
 	yield fork(logoutSaga)
 	yield fork(newDeckSaga)
 	yield fork(matchmakingSaga)
+	yield fork(fbdbSaga)
```
martinkadlec0: If you fork it here instead of at rootSaga, your fbdb saga will be killed and run again every time user logsout and logins again. I assume that is intened?

martinkadlec0 (changes requested):

AndyNovo (commented):
Removed "STATS" switch case in sessions

martinkadlec0:
Closing since it was merged as part of the other refactoring revision.

Issue #21, opened 2023-02-13
After a hermit is knocked out player can add a new hermit on board to make it active and attack with it on the same turn


martinkadlec0:
You should be allowed to activate and attack with a hermit that was already on the board but if you add a new one and make it active you shouldn't be able to attack on the same turn.

xeapnea:
also if you add a hermit after one gets knocked out, the hermit you added is auto selected as the new active hermit (maybe this is a seperate bug?)

PR #61, opened 2023-02-19
Add conditions to check if you can ADD_HERMIT

Sense101:
Fixes https://github.com/martinkadlec0/hc-tcg/issues/21

### DIFF
--- a/server/routines/game.js
+++ b/server/routines/game.js
@@ -71,11 +71,19 @@ function getAvailableActions(game, derivedState) {
 		return actions
 	}
 
-	// TODO - add more conditions (e.g. you can't change active hermit if there is only one.
-	// or you can't add more hermits if all rows are filled)
-	actions.push('ADD_HERMIT')
+	let hermits = 0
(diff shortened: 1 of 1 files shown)


server/routines/game.js
```diff
@@ -71,11 +71,19 @@ function getAvailableActions(game, derivedState) {
 		return actions
 	}
 
-	// TODO - add more conditions (e.g. you can't change active hermit if there is only one.
-	// or you can't add more hermits if all rows are filled)
-	actions.push('ADD_HERMIT')
+	let hermits = 0
+	for (let i = 0; i < currentPlayer.board.rows.length; i++) {
```
martinkadlec0, line 75: For this I think we can stick to `currentPlayer.board.rows.filter(row => row.hermitCard).length`

martinkadlec0 (changes requested):

server/routines/game.js
Sense101: Fair enough

PR #68, opened 2023-02-25
add dropdown for saved decks

jmlyman424:
Saving and loading your decks can be challenging since there's no way to see your decks from the UI. This PR adds a dropdown that automatically queries your Local Storage for keys starting with "Loadout_". It also provides a delete button to remove decks and overwrite protection in case you save a deck with the name of an already saved deck.

Also, aesthetically, I like being able to see the page header as I'm scrolling down, so I made the header sticky and added a small drop shadow.

I noticed Xarlus had attempted to create a dropdown deck selection ([PR#42](https://github.com/martinkadlec0/hc-tcg/pull/42)) but only after starting the same project myself. Great minds think alike I suppose.


martinkadlec0:
Hey, sorry it takes me so much time too look at this, been bit busy last few days.

Note that the plan is to eventually have a nice UI for deck picking (probably with some kind fo speeddial kind of thing), but a select might be nice step in between.

jmlyman424:
No worries! I know you all have been quite busy with some more high-priority stuff. Besides, this PR is mostly a temporary convenience.  I'm currently working on a design for deck building inspired by Hearthstone, but my redux skills are severely limited. I'd like to be able to contribute more but there's a lot I still need to learn.

Looking forward to the future of this project!

client/src/app/deck/deck.tsx
```diff
@@ -86,21 +90,88 @@ const Deck = ({setMenuSection}: Props) => {
...
+		if (!selectedDeck) return console.log('Could not load deck...')
+		const deck: any = localStorage.getItem('Loadout_' + selectedDeck)
 		const deckIds = JSON.parse(deck).filter(
 			(card: CardT) => TYPED_CARDS[card.cardId]
 		)
 		setPickedCards(deckIds)
 	}
+
+	const deleteDeck = (deck: string) => {
+		const confirmDelete = confirm(
```
martinkadlec0, line 161: Same thigns as alert, these dialogs are deprecated and generally shouldn't be use. For the confirm we can use custom model (e.g. similar as for the Imports/Exports)

client/src/app/deck/deck.tsx
```diff
@@ -132,8 +203,29 @@ const Deck = ({setMenuSection}: Props) => {
 					<button type="button" onClick={saveDeck}>
 						Save
 					</button>
-					<button type="button" onClick={loadDeck}>
-						Load
+
+					<select
```
martinkadlec0, line 207: Now that we also have the imports/Exports there isn't enough space to fit everything in on one line. Not sure how to handle that.

martinkadlec0 (changes requested):

martinkadlec0:
Hey, to solve the issue with with not enough space I ended up modifying your PR to use datalist instead. Until we get the new deck screen that should be hopefully good enough:
https://github.com/martinkadlec0/hc-tcg/commit/dd1b032288a30df78a89436e617d18d378edc343

PR #72, opened 2023-02-26
Increase game time out from one hour to two (matching comment)

RyanDMott:


martinkadlec0:
Is this truly needed? Average game takes less than 7 minutes and we have also limited memory/processing power so leaving abandoned games running for too long isn't ideal.

RyanDMott:
Yeah, waiting twice as long to clean up abandoned games would double the memory those games use up.

Now, a seven-minute average game surprises me. Hermits' games regularly go over an hour. I hit the one-hour timeout on my second game. 

Perhaps most players are speedrunners, but perhaps that average also includes a lot of incomplete games? If you can check the percentage of games that take between 30 and 59 minutes, that would clarify how many active games are at risk of timing out.

After more pressing features have been implemented, the best of both worlds might be a popup saying, "It's your turn, but you haven't done anything in five minutes. Please click here if you are still playing. Otherwise we'll assume you left and forfeit the game in 4:59..." That lets active players think about their turns, but kicks abandoned games up to six times faster.

In the meantime, displaying the stop time somewhere would prevent most of the unpleasant surprises.

RyanDMott:
By the way, thank you so much for making and sharing this! It's a really nice implementation and loads of fun.

RyanDMott:
#73 is another factor in the timing question - it appears some players try to force a forfeit by just waiting.

(Now if you like five-minute rounds, but you're playing me, I deliberate for so long that it will feel to you like I'm trying to force a forfeit!)

martinkadlec0:
I have now added 90s turn timer on the dev. domain: https://hc-tcg-dev.fly.dev/ which together with now the game ending because of decking out makes the game timeout pretty much obsolete.

RyanDMott:
LOL, I guess I'll have to change my playing style!

PR #78, opened 2023-02-27
Stacked cards

RoyaLProg:
Not the cleanest of code I admit BUT it does his jobs, could clean it after !

**IMPORTANT** when using stacked card list I do not use the "<animated.div>" !

Should work with no problem (test needed !)

in CSS : just making each card stack on top of the one above + makign card over 3 desapear (for extra large screen to not have random card seen a the bottom)

In card-list.tsx :

+ adding "stack" prameter defaulted at false
+ creating "StackCardOutput" (does the sane thing as "CardOutput" but with a div over the cards taht goes together)


RoyaLProg:
fuck wrong branch x_x

PR #84, opened 2023-02-28
Rarity Icons As Stars

Jimmy474:


client/src/components/card/effect-card-svg.tsx
```diff
@@ -6,6 +6,7 @@ export type EffectCardProps = {
 }
 
 const HermitCard = ({card}: EffectCardProps) => {
+	console.log(document.title);
```
martinkadlec0, line 9: forgotten console log

client/src/components/card/effect-card-svg.tsx
```diff
@@ -48,6 +49,16 @@ const HermitCard = ({card}: EffectCardProps) => {
 					EFFECT
 				</text>
 			</g>
+			{['rare', 'ultra_rare'].includes(card.rarity) ? (
```
martinkadlec0, line 52: This will now appear even in the game. Honestly I feel like the stars don't look that nice on effects/item so I'd prefer to use them only on the deck building screen.

client/src/components/card/effect-card-svg.module.css
```diff
@@ -23,6 +23,10 @@
 	fill: rgb(226, 202, 139);
 }
 
+.rarity {
```
martinkadlec0, line 26: Can you please rebase on latst dev so you don't have the star image change in another commit?

package.json
```diff
@@ -66,5 +65,8 @@
 			"last 1 firefox version",
 			"last 1 safari version"
 		]
+	},
+	"devDependencies": {
+		"vite": "^4.1.4"
```
martinkadlec0, line 70: Why move this one dependency specifically? Anyway, I have intentionally avoided use devDependencies to keep things simple.

martinkadlec0 (changes requested):

package.json
Jimmy474: it auto did that in order to run the game locally

client/src/components/card/effect-card-svg.tsx
Jimmy474: thats why i had console log to check if the title is different so i can only show stars on deck screen
do you know any specific thing that says you are in deck building screen??

PR #86, opened 2023-03-01
Update and improve UI

Sense101:


client/src/app/app.module.css
```diff
@@ -0,0 +1,43 @@
+button {
```
martinkadlec0, line 1: global styles should be in client/src/index.css not here.

client/src/app/app.module.css
```diff
@@ -0,0 +1,43 @@
...
+input {
+	border: 0.3vh solid #000;
+}
+input:focus-visible {
+	outline-offset: 0;
+	outline: 0.5vh solid #333;
+	border-radius: 0.5vh;
+}
+
+::-webkit-scrollbar {
```
martinkadlec0, line 21: These will not fully work in FF. It is okay to have not have it FF but we should check we it loooks okay-ish.

martinkadlec0 (changes requested):

client/src/app/app.module.css
Sense101: Done.

client/src/app/app.module.css
Sense101: I've tested it on firefox, can confirm that the scrollbar stays as the default on firefox. However, it doesn't cause any issues, and the default scrollbar actually looks quite nice.

PR #89, opened 2023-03-01
Deck building and shop?

iAmTotalyAHuman:
Could we have it like on the HC server where you start with a starter deck and have to buy more to customize? It is really fun to have to find rare cards from the pack and not just choose from all the cards. Instead of making a deck based off of your best cards you just have access to everything. It would also be cool if you could sign in and it keeps your saved decks or your deck that you have collected.


martinkadlec0:
@iAmTotalyAHuman why have you asked using a Pull Request?

Anyway, anything that requires a database isn't currently a priority. and I personally don't plan to work on something like that, but someone in the community might pick it up later on.

PR #90, opened 2023-03-01
Make the Logo's subtext configured and update css name.

Maescool:
The logo's subtext can be now changed from config. Updated the css name from notOfficial to logoSubText

### DIFF
--- a/client/src/components/tcg-logo/tcg-logo.tsx
+++ b/client/src/components/tcg-logo/tcg-logo.tsx
@@ -1,12 +1,14 @@
 import css from './tcg-logo.module.css'
+// @ts-ignore
+import {CONFIG} from '../../../../config'
 
 function TcgLogo() {
 	return (
 		/* Logo Container */
 		<div className={css.logo}>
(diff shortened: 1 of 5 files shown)


client/src/components/tcg-logo/tcg-logo.tsx
```diff
@@ -1,12 +1,14 @@
 import css from './tcg-logo.module.css'
+// @ts-ignore
+import {CONFIG} from '../../../../config'
```
martinkadlec0, line 3: We should avoid importing config directly to frontend, can I ask you to use environment variable using the `define` config in `vite.config.js` (also you will need to add type in `global.d.ts`). We already to the same for PORT & LIMITS.

martinkadlec0 (changes requested):

client/src/components/tcg-logo/tcg-logo.tsx
Maescool: Thanks for the advice
Applied in 6dd2cef

PR #96, opened 2023-03-03
Added docker compose file

eyduh:
Added a simple docker compose file.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,12 @@
+version: "2.1"
+services:
+  hctcg:
+    build: .
+    container_name: hc-tcg
+    ports:
+      - 9000:9000
+    volumes:
+      - hc-tcg:/app
+    restart: unless-stopped
+volumes:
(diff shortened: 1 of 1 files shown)


martinkadlec0:
Hey, can I ask what is the motivation for adding this file? I am a docker noob. A quick google search told me this file is used when you run multiple docker containers which is not the case here I think? Is there some other use?

eyduh:
Docker compose allows for easier configuration of docker containers running as services since it makes a docker run command into a markup file. 

This way, if one has multiple containers running on a host there can be an easier overview, rather than keeping track of some command/s one ran at some point, or having to write system services to run these commands at startup etc.

The equivalent in this case would be something like
`docker build . -t hc-tcg:latest && docker run --name hc-tcg -d -p 9000:9000 -v hc-tcg:/app hc-tcg:latest`
and `docker stop hc-tcg` 
if I want to change the port or anything I have to remember that command and what to change.

PR #112, opened 2023-03-09
Added Link Icons Background

Jimmy474:
added background to link icons to match the buttons


martinkadlec0:
Sorry, while there might be something we can do to improve the look of the buttons, giving the button background is not the direction we want to go as it gives too much attention to them.

PR #114, opened 2023-03-11
Deckurl

AndyNovo:
This one introduces a deep link for decks: `https://tcg.prof.ninja/?deck=AQEJCQ4VGR0dHR4fHyQrKy0tLS4uLi8xMTExMTExMjIyWVlZWlpkZGRm` is aiming to load the deck associated with that hash.

If you are not signed in it stays in the url, once at the main page it sends you to the deck screen, once at the deck screen it opens the import/export modal, once there it loads the hash and imports the deck.  It ALSO clears the url parameter so you can go about navigating now.


martinkadlec0:
I've reworked the PR a bit so you get deck on login instead of having to reuse the deck building UI. It was still useful to have the URL logic to copy though :)

https://github.com/martinkadlec0/hc-tcg/commit/9601380816c389641b5fa1745ede20b95e759bf4

PR #130, opened 2023-04-04
Fix bedinfo for consecutive bed use

NikoUY:
bedInfo was being set to True even when the bed was discarded which prevented consecutive use of the bed.

### DIFF
--- a/server/cards/card-plugins/effects/bed.js
+++ b/server/cards/card-plugins/effects/bed.js
@@ -43,8 +43,7 @@ class BedEffectCard extends EffectCard {
 					const hasBed = row.effectCard?.cardId === this.id
 					if (!isSleeping && hasBed) {
 						discardCard(game, row.effectCard)
-					}
-					if (hasBed) bedInfo[rowIndex] = true
+					} else if (hasBed) bedInfo[rowIndex] = true
 				})
 				if (Object.keys(bedInfo).length > 0) {
 					playerState.custom[this.id] = bedInfo


Sense101 (changes requested):
This code does nothing

nikod:
This fixes the case where your bed disappeared at the start of the turn and you added a new bed on the same turn.

Previously bedInfo was set to True when the card was discarded.
When actionEnd was called back hadBed (bedInfo) was True.
Because hadBed was True you would not be put to sleep.
Next turn the bed disappears when turnStart is called because you are not sleeping.

I tested it again and the change works, make sure to restart the server if you already tested it, if you edit the file with the server running the change won’t work, the page reloads but it won’t work, I’m probably being dumb because I have zero web dev experience, I just followed the logic and tested the change.

NikoUY:
This fixes the case where your bed disappeared at the start of the turn and you added a new bed on the same turn.

Previously bedInfo was set to True when the card was discarded.
When actionEnd was called back hadBed (bedInfo) was True.
Because hadBed was True you would not be put to sleep.
Next turn the bed disappears when turnStart is called because you are not sleeping.

I tested it again and the change works, make sure to restart the server if you already tested it, if you edit the file with the server running the change won’t work, the page reloads but it won’t work, I’m probably being dumb because I have zero web dev experience, I just followed the logic and tested the change.

Sense101:
You're missing the point - your else if statement does exactly the same things as the if statement

Sense101:
image
... look at it - the else if statement is exactly the same as the if - meaning nothing is changing in the flow of logic. Please explain.

NikoUY:
I don't know how else to put it, the only thing I can think of is to write some pseudo code for that case

Previously 
If not sleeping and has bed
   discard card
If has bed 
   set bedInfo to true

Now
If not sleeping and has bed 
   discard card
otherwise if sleeping (that's what fails in the first if) and has bed
  set bedInfo

What I proposed in the previous comments to make it more clear and it's what it's basically happening anyway
if not sleeping and has bed
   discard card
   set bedInfo to false
otherwise
   set bedInfo to true

Sense101:
Looks good, thanks a lot. This likely won't make it into the game till we release a new update, but when it is, it'll be credited to you on github and discord.

PR #139, opened 2023-05-02
Fix #136

### LINKED ISSUE #136: [Bug] Bed attached while sleeping doesn't restore health
Steps to reproduce:

1. Attach a bed to your active hermit
2. Get attacked by the opponent, who has no effect card attached
3. Use the Emerald effect card, which moves the bed to the opponent and frees your effect slot
4. Attach another bed to your hermit
5. Your hermit's health is not restored

It should be restored because the card description says it heals whenever it is applied, nothing about having to not be asleep.

I forgot to note if sleep duration resets and I haven't checked if the same happens if you've had your bed removed by the Curse of Vanishing effect card or rare Grian by your opponent, I'll update this if I have a chance to test it

NikoUY:
This fixes that issue and also makes it so the counter resets every time 2 beds are swapped between hermits using an emerald.

### DIFF
--- a/server/cards/card-plugins/effects/bed.js
+++ b/server/cards/card-plugins/effects/bed.js
@@ -34,7 +34,7 @@ class BedEffectCard extends EffectCard {
 		game.hooks.turnStart.tap(this.id, () => {
 			const {currentPlayer, opponentPlayer} = game.ds
 
-			// Need to know which row had bed at start of the turn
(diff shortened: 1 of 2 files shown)


Sense101:
This won't do unfortunately - we can't have cards that specifically reference others, it needs to be more of an abstract system. See if you can set this up without that

PR #150, opened 2023-05-17
Thorns II and Thorns III

NikoUY:
We need new icons.


Sense101:
will wait on new icons before merging tihs, collab with @DevChimera

PR #153, opened 2023-05-18
Changes to the rarity icons

Benji42:
The rarity icons have been changed to be SVG files, and to be higher contrast/more distinct at small sizes. In addition, netherite rank has been renamed to obsidian rank.


Sense101:
please fix linting errors, remove svg files, and unify size of all png images

Benji42:
The issues mentioned should now be fixed.

PR #165, opened 2023-05-23
Hermit palettes

Benji42:
This PR should allow alternate palettes to be defined for cards where they're necessary. The already-implemented alter ego cards have also had their palettes set properly. In the future, palettes can be added for effect & item cards as well, but none are implemented right now (as no alternates exist yet).

Right now, the "alter ego" palette and "pharaoh" palette have been defined.


Sense101:
seems to be slight inconsistency here - some cards like llamadad are still using flags rather than the new functions

Benji42:
Everything should be fixed.

Benji42:
Should be working properly now.

PR #166, opened 2023-05-25
Add additional api functionality

Tyrannicodin:
Added a `gameEndWebhook` to api.js that uses a fetch request to send it to the bot at the url in server-config.json.
Also, the `getOpponentId` function in index.js always returned undefined as it got `playerId` not `id` which is now fixed.


Tyrannicodin:
I've now added a createGame endpoint that allows you to give two playerIds and creates a custom game that the players are immediately added to

PR #167, opened 2023-05-26
Fix lead

NikoUY:
Reverted the pick requirement changes I made for the enederpearl, it's disabled anyways and this is obsolete after the refactor is done.


Sense101:
This is good, but please rebase onto dev

PR #174, opened 2023-05-30
Fixes for Effect and Single Use Cards

NikoUY:


common/cards/card-plugins/effects/_effect-card.js
```diff
@@ -3,28 +3,35 @@ import Card from '../_card'
...
+		if (!defs.description) {
 			throw new Error('Invalid card definition!')
 		}
+
 		/** @type {string} */
-		this.description = description
+		this.description = defs.description
+
+		/** @type {number | undefined} */
+		this.protection = defs.protection
```
Sense101, line 34: Do we really want this here? Or do we just use the methods

common/cards/card-plugins/effects/_effect-card.js
```diff
@@ -3,28 +3,35 @@ import Card from '../_card'
...
+ * @typedef {import('common/types/cards').CardDefs} CardDefs
  * @typedef {import('common/types/cards').CardTypeT} CardTypeT
  * @typedef {import('models/attack-model').AttackResult} AttackResult
  * @typedef {import('utils').GameModel} GameModel
  */
 
 class EffectCard extends Card {
 	/**
-	 * @param {EffectDefs} defs
+	 * @param {EffectDefs & CardDefs} defs
```
Sense101, line 14: CardDefs should be included in EffectDefs, in the ts code. Like EffectDefs = CardDefs & {...

common/cards/card-plugins/hermits/_hermit-card.js
```diff
@@ -59,7 +59,7 @@ class HermitCard extends Card {
 	 * @param {AttackModel} attack
 	 */
 	onAttack(game, instance, attack) {
-		// default is do nothing
+		attack.addDamage(attack.type === 'primary' ? this.primary.damage : this.secondary.damage)
```
Sense101, line 62: remember there can also be zero attacks, meaning that we might add no damage.

common/cards/card-plugins/single-use/_single-use-card.js
```diff
@@ -2,20 +2,37 @@ import {AttackModel} from '../../../../server/models/attack-model'
...
  * @typedef {import('models/attack-model').AttackResult} AttackResult
+ * @typedef {import('utils').GameModel} GameModel
+ * @typedef {import('common/types/cards').DamageT} DamageT
  */
 
+
 class SingleUseCard extends Card {
+	/**
+	* @param {SingleUseDefs & CardDefs} defs
+	*/
```
Sense101, line 17: As above, CardDefs should be included in SingleUseDefs in the ts code

common/cards/card-plugins/single-use/_single-use-card.js
```diff
@@ -2,20 +2,37 @@ import {AttackModel} from '../../../../server/models/attack-model'
...
 
 		if (!defs.description) {
 			throw new Error('Invalid card definition')
 		}
+		/** @type {string} */
 		this.description = defs.description
 
-		this.attachReq = {target: 'player', type: ['single_use']}
+		/** @type {DamageT | undefined} */
+		this.damage = defs.damage
```
Sense101, line 35: Again, this is really not needed, but helps the client to preview the damage. Is there a better way to preview?

common/types/cards.ts
```diff
@@ -49,17 +37,20 @@ export type HermitAttackT = {
 }
 
 export type CardDefs = {
-	type: CardTypeT
+	type?: CardTypeT
```
Sense101, line 40: Nice solution here to allow effect and single use defs to include the base card defs. That should be added here in the ts code.

Sense101 (changes requested):

common/types/cards.ts
Sense101: Actually this doesn't work so well, because each card does need to have a type - it can't be optional, even if it's not passed into the defs for different card types

common/cards/card-plugins/hermits/_hermit-card.js
NikoUY: The cards that use Zero attack are going to override onAttack, this is just for common hermits.

common/types/cards.ts
NikoUY: We can go back to using their own defs types instead of combining them.

PR #182, opened 2023-06-03
Adjacent Pick Requirement

NikoUY:
I also renamed pickedCards to pickedSlots which is more accurate, I should have probably committed that into a different commit but I got carried away while doing some cleanup and working on multiple things at the same time while switching branches, I can change it back later if needed.

getIsRemovable can use GameModel because it's called on validPick which is called client side (server side too).


NikoUY:
I just realized something...

PR #183, opened 2023-06-03
Piston

NikoUY:
Needs #182 to work.


Sense101:
couple notes on this - first, you don't need the @returns jsdoc on inherited methods. Second, do we want to only allow piston to "push" item cards to adjacent empty slots, rather than any item slot?

NikoUY:
That’s what the description says, it also allows you to use another single use card that turn, I think it’s a good trade off.

NikoUY:
I updated the card

Sense101:
please update canAttach

NikoUY:
Should be done.

PR #184, opened 2023-06-03
Ladder

NikoUY:
Needs #182 to work.


Sense101:
I changed onApply so it no longer returns done or invalid, as it's now more universal for all cards.

Sense101:
for clarity on the canAttach return options, please see the original method - you should be returning invalid if the slot doesn't match

NikoUY:
I think it's good now

Sense101:
please update canAttach

NikoUY:
Done

PR #185, opened 2023-06-03
Mumbo Jumbo Rare

Benji42:
Complete


Sense101:
this should not assume the card using the ability is a prankster card, as maybe ren etc will use this attack. Also, please remove the extra 'attack' string from the getInstance key, as this is no longer the default

Benji42:
👍 Is fixed now, thank you!

PR #188, opened 2023-06-03
Wolf

Tyrannicodin:


Sense101:
you need to remove the hooks again during onDetach - Also, you should be hooking into when the opponent attacks.

I also think we could do with a better system for excluding certain cards code from being run... maybe a method?

Sense101:
Looks good now - the _last_ thing now is the attack.shouldIgnoreCard method should not always be returning true, but instead checking if the instance passed in is one of the 4 armor cards, and only then ignore it. See xb rare for an example

PR #189, opened 2023-06-04
ImpulseSV rare

Benji42:
Requires #181 to work.


Sense101:
please remove 'attack' from the instance key method, and make sure you also check that the hermits are afk

Benji42:
I don't believe AFK-ness needs to be checked. To have the card on the board while using this attack, they must be AFK.

PR #192, opened 2023-06-04
Lava bucket

Tyrannicodin:


common/cards/card-plugins/single-use/lava-bucket.js
```diff
@@ -14,24 +14,28 @@ class LavaBucketSingleUseCard extends SingleUseCard {
...
-					['fire', 'poison'].includes(a.id)
-				)
-				if (!hasWaterBucket && !hasDamageEffect) {
-					opponentActiveRow.ailments.push({id: 'fire', duration: -1})
-				}
-				return 'DONE'
-			}
+	onApply(game, instance, pos, pickedSlots) {
+		const opponentActiveRow = pos.otherPlayer.board.activeRow
+		if (opponentActiveRow === null) return 'INVALID'
```
NikoUY, line 25: We don't return INVALID/DONE anymore

NikoUY (changes requested):
It's missing canApply and you need to update the description to the new one on the spreadsheet/

PR #193, opened 2023-06-04
Turtle shell

Tyrannicodin:


Sense101:
please fix to use the new hooks, thanks

Tyrannicodin:
Not completely sure what you mean by that but I have updated to properly cleanup. A potential error I can see cropping up is that if the player uses fire charge and discards the turtle shell it might not properly delete the hooks.

common/cards/card-plugins/effects/turtle-shell.js
```diff
@@ -39,6 +37,76 @@ class TurtleShellEffectCard extends EffectCard {
...
+		return 'YES'
+	}
+
+	/**
+	 * @param {GameModel} game
+	 * @param {string} instance
+	 * @param {import('../../../types/cards').CardPos} pos
+	 */
+	onAttach(game, instance, pos) {
+		const instanceKey = this.getKey(instance)
```
Sense101, line 68: this should be getInstanceKey

common/cards/card-plugins/effects/turtle-shell.js
```diff
@@ -39,6 +37,76 @@ class TurtleShellEffectCard extends EffectCard {
...
+			if (
+				pos.player.custom[instanceKey] === true &&
+				attack.target.row.effectCard?.cardInstance === instance
+			) {
+				attack.multiplyDamage(0).lockDamage()
+			}
+			return attack
+		}
+
+		pos.otherPlayer.hooks.afterAttack[instance] = () => {
```
Sense101, line 81: this will get discarded before any nextAttacks are carried out... this is a me problem, and I will change it so afterAttack is always after *all* attacks

common/cards/card-plugins/effects/turtle-shell.js
```diff
@@ -39,6 +37,76 @@ class TurtleShellEffectCard extends EffectCard {
...
+			if (pos.player.custom[instanceKey] === true) {
+				discardCard(game, {cardId: this.id, cardInstance: instance})
+			}
+		}
+
+		pos.otherPlayer.hooks.turnStart[instance] = () => {
```
Sense101, line 87: I see a slight problem with this that means we may need another hook - what if at the start of your turn you have no active hermit because you just lost a life, and then you switch to our hermit?

PR #194, opened 2023-06-04
Mending

Tyrannicodin:


Sense101:
make sure to change to new mending as per spreadsheet, plus remove hooks on detach

NikoUY (changes requested):
This card is outdated, you need to update it using the new description.

common/cards/card-plugins/single-use/mending.js
```diff
@@ -0,0 +1,63 @@
...
+		this.pickReqs = /** @satisfies {Array<PickRequirmentT>} */ ([
+			{target: 'player', type: ['effect'], amount: 1, active: true},
+			{
+				target: 'player',
+				type: ['effect'],
+				amount: 1,
+				empty: true,
+				active: false,
+			},
+		])
```
Sense101, line 31: pickOn and pickReq needs to be passed within super

common/cards/card-plugins/single-use/mending.js
```diff
@@ -0,0 +1,63 @@
...
+	}
+
```
Sense101, line 33: you need the canApply method returning true, so this can be applied

common/cards/card-plugins/single-use/mending.js
```diff
@@ -0,0 +1,63 @@
...
+		super({
+			id: 'mending',
+			name: 'Mending',
+			rarity: 'ultra_rare',
+			description:
+				'Move any attached effect card from your active Hermit to an AFK Hermit.',
+		})
+
+		this.pickOn = 'apply'
+		this.pickReqs = /** @satisfies {Array<PickRequirmentT>} */ ([
```
Sense101, line 22: no need for @ satisfies anymore, remove pls

common/cards/card-plugins/single-use/mending.js
Sense101, line 33: you also need a canAttach method that makes sure the active hermit has an effect card - see one of niko's merged signle use cards.

common/cards/card-plugins/single-use/mending.js
```diff
@@ -0,0 +1,63 @@
+import singleUseCard from './_single-use-card'
+import {GameModel} from '../../../../server/models/game-model'
+import {validPick} from '../../../../server/utils/reqs'
+
+/**
+ * @typedef {import('common/types/pick-process').PickRequirmentT} PickRequirmentT
+ *
+ */
+
+// TODO - Must work with Gemintay ability to use two single use cards per turn (should mend the first one)
```
Sense101, line 10: With the new functionality, this should work fine, so this todo can be removed - provided you've set up canAttach correctly -see below

Sense101 (changes requested):

Sense101 (approved):
Still no canApply here, but merging anyways

PR #199, opened 2023-06-05
Sweeping Edge

NikoUY:


common/cards/card-plugins/single-use/sweeping-edge.js
```diff
@@ -1,6 +1,11 @@
 import SingleUseCard from './_single-use-card'
 import {discardCard} from '../../../../server/utils'
-import {GameModel} from '../../../../server/models/game-model'
```
Sense101, line 3: GameModel is a class, it should not be imported as a typedef

common/cards/card-plugins/single-use/sweeping-edge.js
```diff
@@ -9,35 +14,60 @@ class SweepingEdgeSingleUseCard extends SingleUseCard {
...
-				].filter(Boolean)
-
-				targetRows.forEach((row) => {
-					if (row.effectCard) discardCard(game, row.effectCard)
-				})
-
-				return 'DONE'
-			}
-		})
+	canAttach(game, pos) {
```
Sense101, line 25: Clean up canAttach as specified on discord

NikoUY:
The card should be done.

PR #202, opened 2023-06-05
Llamadad Rare

Benji42:


Sense101:
ondetatch also needs to use new pos system

common/cards/card-plugins/hermits/llamadad-rare.js
```diff
@@ -28,23 +28,35 @@ class LlamadadRareHermitCard extends HermitCard {
 
 	/**
 	 * @param {GameModel} game
+	 * @param {string} instance
```
Sense101, line 31: please add @ param for pos

common/cards/card-plugins/hermits/llamadad-rare.js
```diff
@@ -28,23 +28,35 @@ class LlamadadRareHermitCard extends HermitCard {
...
+				attack.addDamage(40)
+			}
+		}
+	}
 
-			return target
-		})
+	/**
+	 * @param {GameModel} game
+	 * @param {string} instance
```
Sense101, line 54: please add @ param for pos

Sense101 (changes requested):
Other than two minor things, this is good

PR #203, opened 2023-06-06
Totem

Tyrannicodin:


common/cards/card-plugins/effects/totem.js
```diff
@@ -9,31 +8,37 @@ class TotemEffectCard extends EffectCard {
...
-		// attacks
-		game.hooks.attack.tap(this.id, (target) => {
-			if (target.effectCardId === this.id) {
-				target.recovery.push({amount: this.recoverAmount, discardEffect: true})
-			}
-			return target
-		})
-
+	onAttach(game, instance, pos) {
 		game.hooks.hermitDeath.tap(this.id, (recovery, deathInfo) => {
```
NikoUY, line 22: Use the afterAttack hook and this needs to be redone to work with that hook, make sure to discard the totem.

common/cards/card-plugins/effects/totem.js
```diff
@@ -9,31 +8,37 @@ class TotemEffectCard extends EffectCard {
...
+					amount: 10,
+					discardEffect: true,
+				})
+			}
 			return recovery
 		})
 	}
+
+	/**
+	 *
```
NikoUY, line 34: This can be removed

common/cards/card-plugins/effects/totem.js
```diff
@@ -9,30 +8,35 @@ class TotemEffectCard extends EffectCard {
...
-		// attacks
-		game.hooks.attack.tap(this.id, (target) => {
-			if (target.effectCardId === this.id) {
-				target.recovery.push({amount: this.recoverAmount, discardEffect: true})
-			}
-			return target
-		})
+	onAttach(game, instance, pos) {
+		pos.otherPlayer.hooks.afterAttack[instance] = (result) => {
+			if (!result.attack.target.row || result.attack.target.row.health > 0)
```
Sense101, line 22: Please check explicitly that the row health is not undefined. You can also define a "row" variable which is result.attack.target.row, to keep the code cleaner.

Sense101 (changes requested):
One more change., then this can be merged :)

PR #216, opened 2023-06-09
Shield, Netherite, Diamond, Iron Gold and Chainmail Armor 

NikoUY:
I figure that because they basically share a lot of their code I can just create one PR


common/cards/card-plugins/effects/chainmail-armor.js
```diff
@@ -0,0 +1,50 @@
...
+	 * @param {CardPos} pos
+	 */
+	onAttach(game, instance, pos) {
+		const {otherPlayer} = pos
+
+		otherPlayer.hooks.onAttack[instance] = (attack, pickedSlots) => {
+			if (attack.target.index !== pos.rowIndex || attack.type === 'ailment')
+				return
+			if (attack.type === 'effect') {
+				attack.reduceDamage(attack.damage)
```
Sense101, line 30: you can just set the damage to 0 here, but hte important thing is to also lock the damage so it can never be added to

common/cards/card-plugins/effects/chainmail-armor.js
```diff
@@ -0,0 +1,50 @@
...
+	/**
+	 * @param {GameModel} game
+	 * @param {string} instance
+	 * @param {CardPos} pos
+	 */
+	onAttach(game, instance, pos) {
+		const {otherPlayer} = pos
+
+		otherPlayer.hooks.onAttack[instance] = (attack, pickedSlots) => {
+			if (attack.target.index !== pos.rowIndex || attack.type === 'ailment')
```
Sense101, line 27: You don't need to check for ailment if you've got a check right after filtering out anything that isn't an effect attack

common/cards/card-plugins/effects/diamond-armor.js
```diff
@@ -1,30 +1,66 @@
...
 	 */
-	onDefence(game, instance, attack) {
-		if (attack.type !== 'ailment') {
-			attack.defence.damageReduction += 30
+	onAttach(game, instance, pos) {
+		const {otherPlayer, player} = pos
+		const instanceKey = this.getInstanceKey(instance)
+
+		otherPlayer.hooks.onAttack[instance] = (attack, pickedSlots) => {
+			if (attack.target.index !== pos.rowIndex || attack.type === 'ailment')
```
Sense101, line 29: no need to filter out ailment, you are filtering for effects

common/cards/card-plugins/effects/diamond-armor.js
```diff
@@ -1,30 +1,66 @@
...
+				return
+			if (attack.type === 'effect') {
+				attack.reduceDamage(attack.damage)
```
Sense101, line 32: same as above, lock the damage

common/cards/card-plugins/effects/diamond-armor.js
NikoUY: Not needed for the Chainmail Armor but it's needed for the Diamond and Netherite armor, they block up to 20 damage from other sources.

common/cards/card-plugins/effects/chainmail-armor.js
NikoUY: I think it's better to use reduceDamage because of the lock, by setting it directly we are bypassing it.

PR #218, opened 2023-06-10
Knocback ported

NikoUY:


Sense101:
You should be checking if activeRow.health is undefined, too. In addition, please make sure this works correctly with curse of binding, we may need helper method for that. Same issue with tango

NikoUY:
Curse of Binding applies the following turn so we don't have to check.

PR #220, opened 2023-06-10
Bad Omen

NikoUY:
I implemented it to test Fortune, it's not displaying that the effect it's active, not sure how you want to display ailments that affect the player.

### DIFF
--- a/common/types/game-state.ts
+++ b/common/types/game-state.ts
@@ -12,7 +12,7 @@ export type CardT = {
 }
 
 export type Ailment = {
-	id: 'poison' | 'fire' | 'sleeping' | 'knockedout' | 'slowness' | 'bad_omen'
+	id: 'poison' | 'fire' | 'sleeping' | 'knockedout' | 'slowness' | 'badomen'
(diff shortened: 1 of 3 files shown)


common/types/game-state.ts
```diff
@@ -12,12 +12,7 @@ export type CardT = {
 }
 
 export type Ailment = {
-	id: 'poison' | 'fire' | 'sleeping' | 'knockedout' | 'slowness'
-	duration: number
-}
-
-export type PlayerAilment = {
-	id: 'badomen'
+	id: 'poison' | 'fire' | 'sleeping' | 'knockedout' | 'slowness' | 'bad_omen'
```
Sense101, line 15: Inconsistency here - please decide if it's "bad_omen" or "badomen"

Sense101 (changes requested):

common/types/game-state.ts
Sense101: meaning that in the card itself, you've used badomen

PR #223, opened 2023-06-10
Helsknight rare

Benji42:


common/cards/card-plugins/hermits/helsknight-rare.js
```diff
@@ -0,0 +1,84 @@
+import HermitCard from './_hermit-card'
+import {flipCoin, discardSingleUse} from '../../../../server/utils'
+import {GameModel} from '../../../../server/models/game-model'
+
+// Because of this card we can't rely elsewhere on the suCard to be in state on turnEnd hook
```
Sense101, line 5: Is this comment accurate, or copied from another card?

common/cards/card-plugins/hermits/helsknight-rare.js
```diff
@@ -0,0 +1,84 @@
...
+	 * @param {import('../../../types/cards').CardPos} pos
+	 */
+	onAttach(game, instance, pos) {
+		const {player, otherPlayer} = pos
+		const instanceKey = this.getInstanceKey(instance)
+
+		player.hooks.onAttack[instance] = (attack) => {
+			if (attack.id !== instanceKey || attack.type !== 'secondary') return
+
+			player.custom[this.id] = true
```
Sense101, line 42: don't use id for custom - use the instance key.

common/cards/card-plugins/hermits/helsknight-rare.js
```diff
@@ -0,0 +1,84 @@
...
+		otherPlayer.hooks.afterAttack[instance] = () => {
+			if (
+				player.custom[this.id] == true &&
+				otherPlayer.board.singleUseCardUsed &&
+				otherPlayer.board.singleUseCard
+			) {
+				const coinFlip = flipCoin(otherPlayer)
+				otherPlayer.coinFlips[this.id] = coinFlip
+
+				if (coinFlip[0] == 'heads')
```
Sense101, line 54: This is just semantic, but as this is too long to fit on one line please add {} for this if

Sense101 (commented):
Mostly looks ok - did we confirm how this works with trident?

common/cards/card-plugins/hermits/helsknight-rare.js
```diff
@@ -0,0 +1,91 @@
...
+				cost: ['any'],
+				damage: 40,
+				power: null,
+			},
+			secondary: {
+				name: 'Trap Hole',
+				cost: ['pvp', 'pvp', 'pvp'],
+				damage: 100,
+				power:
+					'If opponent uses a single use effect card on their next turn, they must flip acoin. If heads, you take the card after its effect is applied and add it to your hand.',
```
NikoUY, line 24: you wrote "acoin" instead of "a coin"

NikoUY (changes requested):
This should only apply for 1 turn, right now if the opponent does not attack the effect will stay active, you should use turnOver to get rid of the custom value.

common/cards/card-plugins/hermits/helsknight-rare.js
```diff
@@ -0,0 +1,91 @@
...
+		otherPlayer.hooks.onAttack[instance] = () => {
+			if (!player.custom[instance]) return
+			otherPlayer.hooks.afterAttack[instance] = (attackResult) => {
+				if (attackResult.attack.type === 'effect') {
+					if (!otherPlayer.board.singleUseCard) return
+					const coinFlip = flipCoin(otherPlayer, this.id, 1)
+					otherPlayer.coinFlips[this.id] = coinFlip
+
+					if (coinFlip[0] == 'heads') {
+						player.hand.push(otherPlayer.board.singleUseCard)
```
NikoUY, line 56: onDetach and the onDetach hooks should be called before removing the single use card, you can use moveCardToHand, I added it on #264.

NikoUY (changes requested):

PR #225, opened 2023-06-11
Lightning Rod and Target Block

NikoUY:


common/cards/card-plugins/effects/lightning-rod.js
```diff
@@ -0,0 +1,72 @@
...
+			description:
+				"Attach to any of your active or AFK Hermits.\n\n All damage done to you on your opponent's next turn is taken by the Hermit this card is attached to.\n\nDiscard after damage is taken.",
+		})
+	}
+
+	/**
+	 * @param {GameModel} game
+	 * @param {CardPos} pos
+	 */
+	canAttach(game, pos) {
```
Sense101, line 25: Why can't this be removed and the default used? isn't it identical? Everything else here looks good.

Sense101 (changes requested):

common/cards/card-plugins/effects/lightning-rod.js
NikoUY: It can be removed, I removed the restriction but I didn’t realize that without it this does the same as the default.

PR #228, opened 2023-06-11
Renbob rare hermit

Benji42:
Requires #217


Sense101 (changes requested):
Not a fan of the whole getIsRedirecting thing, as we discussed - can this please get removed here

Benji42:
Requires #272, but works now

PR #231, opened 2023-06-11
Update deck urls embedded in urls for the new system

Tyrannicodin:
This fixes the issue where the server would crash whenever someone tried to login with a deck in their url.


Tyrannicodin:
Whoops, currently doesn't build as I double negated when unnecessary

PR #232, opened 2023-06-11
Improve asset quality for base hermits

Benji42:
Images of regular hermits were swapped out for higher quality versions provided by the artist. Bdoubleo was switched out to Hoffen's render, instead of the previous version.


Benji42:
This should now be complete. I'll make sure the alter egos hermits are sized properly in a future PR, as those aren't all done yet.

Benji42:
Should be perfect now!

PR #235, opened 2023-06-13
Golden Axe

NikoUY:


Sense101 (changes requested):
Just gonna point out - this new implementation is wayyy worse than the last, which is not our fault, but I might change it back to the old implementation.

However, as it is at the moment, this needs to be using the new power system, rather than checking if the item card is rare.

NikoUY:
Any reason why getEnergy needs game, instance and CardPos?, that seems kinda overkill, anyways, it should be done.

server/utils/index.js
```diff
@@ -357,3 +367,24 @@ export function rowHasItem(row) {
...
+	const itemCards = row.itemCards
+	let total = 0
+	for (const itemCard of itemCards) {
+		if (!itemCard) continue
+		const cardInfo = ITEM_CARDS[itemCard.cardId]
+		// String
+		if (!cardInfo) continue
+		const pos = getCardPos(game, itemCard.cardInstance)
+		if (!pos) continue
+		total += cardInfo.getEnergy(game, itemCard.cardInstance, pos)
```
Sense101, line 386: how are you adding an array to a number here??

Sense101 (approved):
This is good now, but please enable a linter for vscode, so you get errors for things like that.

PR #236, opened 2023-06-13
Scar rare hermit

Benji42:
Unlike my other PRs, I'm not entirely sure this works properly. It seems to be working now, but it's possible I made a mistake.


common/cards/card-plugins/hermits/goodtimeswithscar-rare.js
```diff
@@ -26,106 +26,58 @@ class GoodTimesWithScarRareHermitCard extends HermitCard {
...
+				attack.id !== this.getInstanceKey(instance) ||
+				attack.type !== 'secondary'
+			)
+				return
+
+			player.custom[reviveNextTurn] = true
+		}
+
+		otherPlayer.hooks.afterAttack[instance] = () => {
+			if (player.board.activeRow !== rowIndex) return
```
Sense101, line 55: You should be checking the target of the attack here, not the active row

common/cards/card-plugins/hermits/goodtimeswithscar-rare.js
```diff
@@ -26,106 +26,58 @@ class GoodTimesWithScarRareHermitCard extends HermitCard {
...
+			if (player.custom[reviveNextTurn] && !player.custom[scarRevivedKey]) {
+				if (row?.health !== 0) return
```
Sense101, line 57: It's worth checking to see that a) it's not undefined and b) it's less than or equal to 0

Sense101 (changes requested):

Sense101 (approved):
Slight overkill with the number of checks now, but I'm just going to merge this and remove the extras myself. Great job!

PR #240, opened 2023-06-14
Tango

NikoUY:
Needs #239


Sense101:
you need to check if curse of binding has been used here - we need a nice way to do that, possibly a helper method, and in a way that does not mention any cards by id

PR #255, opened 2023-06-16
Grian

NikoUY:
Needs #259 and #264


NikoUY:
I realized that we need to a helper for moving cards around otherwise onDetach and onAttach don't get called, I'll work on that tomorrow.

PR #261, opened 2023-06-17
Evil X

NikoUY:
Merge #259 first


Benji42:
you probably want to remove the image titled "evilx" from this PR, as it's not being used by any hermit

NikoUY:
Thanks, I was using that before I realized that the id was supposed to be evilxisuma_rare

common/cards/card-plugins/hermits/evilxisuma_rare.js
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	/**
+	 * @param {GameModel} game
+	 * @param {string} instance
+	 * @param {CardPos} pos
+	 */
+	onDetach(game, instance, pos) {
+		const {player} = pos
+		delete player.hooks.onAttack[instance]
```
Sense101, line 120: why are you not deleting all the hooks?

Sense101 (changes requested):

common/cards/card-plugins/hermits/evilxisuma_rare.js
NikoUY: Not needed, onFollowUp and onFollowUpTimeout are created inside onAttack and they delete themselves as soon as they are called, onTurnEnd and blockedActions get deleted on the end of the opponents turn, I can’t delete them when Evil X dies from a backslash because the effect would stop early.

PR #263, opened 2023-06-17
Fixed Totem

NikoUY:


Sense101:
Please explain

NikoUY:
Before it was returning when the health was 0

**!**targetRow.health

PR #265, opened 2023-06-17
Hermit shouldn't wake up when the bed is detached

NikoUY:


common/cards/card-plugins/effects/bed.js
```diff
@@ -67,21 +67,6 @@ class BedEffectCard extends EffectCard {
...
 	}
-
-	/**
-	 * @param {GameModel} game
-	 * @param {string} instance
-	 * @param {import('../../../types/cards').CardPos} pos
-	 */
-	onDetach(game, instance, pos) {
-		const {player, row} = pos
-		delete player.hooks.onTurnStart[instance]
```
Sense101, line 78: You should still be deleting the hooks

Sense101 (changes requested):

PR #267, opened 2023-06-17
Zed shouldn't be hit by weakness and effects

NikoUY:


Sense101:
weakness I agree, but effects? should they also be reflected?

NikoUY:
That’s how it works now with the exception of Hypno, we had a discussion a couple weeks ago on discord because Hypno was able to ignore him and we came to the conclusion that he should stop everything.

PR #270, opened 2023-06-18
Slowness Potion

NikoUY:


common/cards/card-plugins/single-use/potion-of-slowness.js
```diff
@@ -9,22 +13,31 @@ class PotionOfSlownessSingleUseCard extends SingleUseCard {
...
+		const {otherPlayer} = pos
+		const opponentActiveRow = getActiveRow(otherPlayer)
+		if (!opponentActiveRow) return
+		opponentActiveRow.ailments.push(
+			{id: 'slowness', duration: 1},
+			{id: 'weakness', duration: 1},
+			{id: 'fire', duration: 1},
+			{id: 'poison', duration: 1},
+			{id: 'badomen', duration: 1},
+			{id: 'sleeping', duration: 1}
```
Sense101, line 39: Looks like you were testing some stuff here.. but we don't want every ailment to come from the weakness potion!

Sense101 (changes requested):

common/cards/card-plugins/single-use/potion-of-slowness.js
NikoUY: oops, yes, I was testing the status icons and I forgot to undo it, may bad.

PR #272, opened 2023-06-18
Added support for null targets

Benji42:
With this PR, the target of an attack can be null. This is needed for both Renbob and Human Cleo.


Sense101:
Are you sure this is needed? I'm really ont a fan of this, the whole idea is that you have to have an attack target

NikoUY:
This is for Renbob+Target Block/Lightning Rod, basically he can use it now without getIsRedirecting, the attack is created without a target and gets redirected to a target by a target block/lightning rod, if they were not played the attack does nothing. 

It’s also useful but not necessarily needed for Human Cleo, because of the way she works it means that the opponent needs to pick after they choose an attack, this is accomplished by deleting the target on beforeAttack, adding a followUp so they can choose a new target and then creating a new attack equal to the original, alternatively the attack would need to be reduced to 0 but then you have to worry about backslash attack, this just happens to make it a lot easier.

Sense101:
Changed branch to beta, I feel bad.. but can you resolve conflicts?

Sense101:
Also can you confirm that every card is properly checking that target isn't null

PR #287, opened 2023-06-19
Fixed Target Block and Lightning Rod redirecting backlash and ailment attacks.

NikoUY:


Sense101 (approved):
You left a console.log here, but I'm just going to remove it

PR #293, opened 2023-06-19
Temporary token costs

Benji42:
This goal of this PR is to add a token cost in the right ballpark to make balancing much easier. Right now, the beta server is filled with people abusing the 0 token cost new cards and creating decks that don't actually reflect how the new cards will be used. This should give a baseline to start from.


Sense101:
can we get these reorganized so the alter ego cards are below the rest?

Benji42:
NP, just reorganized it

PR #302, opened 2023-06-20
Fix Hotguy not deleting instance key at the end of turn

Benji42:


Sense101 (changes requested):
For player.custom, you should be using the instance key, not the instance

Sense101 (approved):

PR #315, opened 2023-06-23
Improved flipCoin

NikoUY:
- Fortune stays active until the start of your next turn
- Coin flips get automatically added to the player state
- The same card can toss a coin multiple times per turn and they will be displayed
- Since fortune stays active until the end of your next turn you can use flipCoin whenever you want as long as you use the correct arguments.

This PR will probably conflict with #313 , I'll fix the PR that creates a conflict after one of them gets merged.

PR #322, opened 2023-06-23
Fixed ailments lasting 1 turn less than they should

NikoUY:
Originally I moved the aliment logic back to where it was before but then I realized that because of the changes to flipCoin it means that Bad Omen can't stay active until the start of your turn which meant that if you tossed a coin Bad Omen was staying active when it shouldn't, that means that if I wanted it to work properly I needed to change the bed to be removed at the end of your turn otherwise the hermit would have a bed without sleeping at the end of their turn.

One thing to keep in mind is that because the aliment logic was moved to the end of the turn you need to set the aliment to last 1 extra turn if you are applying an ailment to the current player.


NikoUY:
Fixed the issue with the linter on #315

PR #323, opened 2023-06-25
Armor Stand

NikoUY:
Implemented it as an Effect Card, it works already but it has a few caveats.
- Effect Cards don't have health which is fine because I can set the health of the row but I can't heal it because it doesn't have a max health, I could add an exception to the healing cards, make it so it can be healed without restrictions or just say that it can't be healed at all.
- I had to improve the pick system to be able to differentiate between slots and cards, otherwise the ladder would be able to be used with it, the fire charge would be able to remove it, etc... basically the old code was mixing slots and cards, now that I implemented that we should probably decide what cards intended to work with hermits should work with this card if any and the same for cards that works with effect cards.
- I've a screenshot of the card on the hermitcraft server that says that the card attaches, not sure what that means but for now I copy/pasted the shield code into this card, I was thinking that maybe instead of working like a shield it could increase the health by 50 (going over the base health of the hermit card) and it breaks when the hermit gets attacked, it basically works like a shield but if the opponent removes it it won't remove the extra health, it also means that a card can't be attached until it breaks.


NikoUY:
Should be ready to merge.

PR #354, opened 2023-07-10
Transformed CardPos into a Class

NikoUY:
Fixes cards using the wrong position when moved by the Ladder/EnderPearl

Everything that was using pos.row, pos.rowIndex and pos.slot was broken after it was moved by those cards, this fixes it by updating them on the class if the position changed, detaching and attaching the cards again wasn't an option because they would lose their current state (custom/hooks).

Right now I'm sending the CardPos to canAttach without an instance and then setting it afterwards, otherwise it would trigger the check/search but the card won't be there so rowIndex, etc won't return anything even though the correct info was stored inside the class, basically what I did was to skip the instance check inside CardPos until the instance is set, we may want to consider using SlotPos (needs to be changed a bit) for canAttach instead.  
Also merging PickedSlotT and SlotPos may not be a bad idea, I just didn't have enough time and this was good enough for now.

Removed getCardAtPos, I'm storing the instance inside CardPos I can just give you the card using pos.card.

I tested a few cards with this but I don't have a lot of time right now, I would appreciate if someone can do a bit more testing, basically search whatever is using isTargetingPos/other helpers that use CardPos or anything that uses rowIndex/row inside the hooks (slot is not necessary because it won't change atm), those are the things that may break.

A few things to keep in mind.
- pos.slot can be null otherwise if for whatever reason the card can't be found there needs to be a way for whatever is using CardPos to know if it's using a valid position.
- You need to get rowIndex/row/slot inside the hooks otherwise you won't get the correct info after the card was moved, I changed the cards that were doing that.

### DIFF
--- a/common/cards/card-plugins/_card.js
+++ b/common/cards/card-plugins/_card.js
@@ -1,10 +1,10 @@
 // TODO - more validations (types, rarirites, other fields, ...)
 
 import {GameModel} from '../../../server/models/game-model'
+import CardPos from '../../../server/models/card-pos-model'
 
 /**
  * @typedef {import("../../types/cards").CardDefs} CardDefs
- * @typedef {import("../../types/cards").CardPos} CardPos
  * @typedef {import("../../types/cards").CardTypeT} CardTypeT
  * @typedef {import("../../types/cards").CardRarityT} CardRarityT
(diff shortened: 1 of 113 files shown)


NikoUY:
I won't be available for a few days so feel free to do whatever you want with this and the Armor Stand PR.

PR #366, opened 2023-07-30
Fix string not being attachable

Benji42:


Sense101:
This is definitely not the correct solution, as the canAttachToCard method is on all cards - I'm hoping my changes will have fixed this but will confirm.

PR #367, opened 2023-07-30
Allow Armour Stand to be attachable when player has no Hermit cards in hand

Benji42:
The way I did this is I changed `isAttachable()` to be a function that is more universally applicable.

However, `isAttachableToSlotType()` and `canAttach()` need to be updated separately if the slot type differs from the type the parent class is attachable to, but this was already a problem beforehand. I don't see an easy way to fix that at the moment.


Sense101:
I wasn't aware you'd opened a pr for this till now, but I've been working on a much more universal fix for the issue that allows cards to define what play actions become available when they are in the hand.

PR #371, opened 2023-08-05
Fixed placement on row 0 and some effect cards being able to be attached to the Armor Stand

NikoUY:


Sense101 (approved):
Great job on this

PR #376, opened 2023-08-08
Fix chat notification when you chat

Tyrannicodin:


Sense101:
should be called opponent for consistency

PR #390, opened 2023-08-31
Removes the parts of the code which heal armor stand.

Rvtar:
Changed cards include Rare Keralis, Potatoboy, Golden apple, Instant health and their splash versions.


Sense101:
This is no good, as you'll still be able to select armor stand with those cards, but it will do nothing. We need to prevent them from being selected as well, which sin't so simple

Sense101:
Added in latest version, however, this PR was not merged due to the complexity needed to make it work. Thank you for the contribution.

PR #391, opened 2023-09-04
Add onDetach for egg

Tyrannicodin:
Probably missed something, lmk


Sense101:
Took a look, the problem is actually that the funtionality of egg in the afterAttack hook, should not be active until the card has been applied, I will be merging this and then making changes so that that only kicks in after the card is applied.

This means that the card will still work with gem, as even when it's removed for a second effect, the detach will not prevent the apply from triggering the egg's functionality.

PR #394, opened 2023-09-06
Prevent several weakness attacks on healing

Tyrannicodin:


Sense101:
I don't really like how hacky this is, plus you commited the debug config :)

Tyrannicodin:
Fair, any suggestions on a better way to do this? Maybe another attack parameter, or what if 0 damage attacks never created weakness? I had thought of that but not sure if it'd break something.

Tyrannicodin:
New ailments system should fix this

PR #400, opened 2023-11-10
Ren and Cleo modal fixes

Benji42:
Now, Ren and Cleo both have a modal to select which attack to copy. The problem with the hooks being deleted every turn has a also been fixed, so they're only removed when the mimicked card changes.

In addition, X's modal was updated to be more consistent with the other two.

The modal requests have an option to supply a picked slot now, to work with Ren and Cleo, which seemed like the best solution to me. However I'm not totally sure (tbf modal requests are still messy in general)


Sense101:
This ready to merge now?

Benji42:
yes

client/src/logic/game/tasks/action-logic-saga.ts
```diff
@@ -42,18 +42,18 @@ function* actionLogicSaga(gameState: LocalGameState): SagaIterator {
 	const pState = gameState.players[playerId]
 	const lastActionResult = gameState.lastActionResult
 
-	if (gameState.currentModalId !== null) {
-		const id = gameState.currentModalId
+	if (gameState.currentModalData && gameState.currentModalData.modalId) {
+		const id = gameState.currentModalData?.modalId
 		if (id === 'grian_rare') {
 			yield fork(borrowSaga)
```
Sense101, line 48: One more thing, actually - can you see how here a huge portion of this if/else is now meaningless? Unless if it's grian, for which we have a custom saga, it should just be setOpenedModal(id)

Sense101 (changes requested):

client/src/logic/game/tasks/action-logic-saga.ts
Benji42: i'll take a look

Benji42: cool fixed

Sense101 (approved):

PR #404, opened 2023-11-12
Backend changes needed for the new Advent of TCG cards

Benji42:


Benji42:
damn, that's a lot of commits

Sense101:
Just so I'm clear this is the big pr without the actual advent cards? So this includes the new ailment system?

Sense101 (changes requested):
This has not got the fixes @Tyrannicodin added to the other pr, what we need is a pr with those changes, but without the new cards

Sense101 (approved):

PR #406, opened 2023-11-16
Add system to check if expansions have cards associated

Benji42:
Now, expansions will only show in the dropdown menu if they have cards associated with them. This will allow the advent of TCG stuff to be merged into main without having to remove the expansion information from expansions.json every time. We just need to disable cards.


Sense101 (approved):
At some point I would like to rework our folder structure so that each expansion has it's own folder and it's much easier to disable them, however this is a good change for now. Have you confirmed that removing cards after they are already in people's decks will not break the game?

PR #423, opened 2023-12-08
Ldshadowlady: add clarification to description

Benji42:


common/cards/advent-of-tcg/hermits/ldshadowlady-rare.ts
```diff
@@ -23,7 +23,8 @@ class LDShadowLadyRareHermitCard extends HermitCard {
 				name: 'Evict',
 				cost: ['terraform', 'terraform', 'any'],
 				damage: 90,
-				power: 'Opposing active Hermit must move to a free slot if one is available.',
+				power:
+					"If one is available, move your opponent's active Hermit and any attached cards to an open slot on their board.",
```
Sense101, line 27: This is still quite confusing, better to leave the "if one is available" after the rest of the texr

Sense101 (changes requested):

common/cards/advent-of-tcg/hermits/ldshadowlady-rare.ts
Sense101: Changing it myself

PR #444, opened 2024-01-04
Prevent single_use_attack when sleeping

Tyrannicodin:


Sense101:
Should this still be merged?

PR #463, opened 2024-01-28
Fix issues with bad omen interactions with pearl

Benji42:


Sense101 (changes requested):
You've committed the debug config...

Sense101 (approved):

PR #470, opened 2024-02-25
Evil Xisuma (EX) Boss

Screaper91:
Adds the boss fight from HC:S9 world download with voice lines, rules page, and advent card compatibility.

Challenging EX does not contribute to statistics at all since it is a private game.


Sense101:
Hey, would you be able to join the discord and contact us about this there?

Lunarmagpie:
I have finally written the abstractions I wanted to cleanly do games like this so I think its finally time we can clean this up and merge it.

The main thing that I think should be done is using a `PlayerComponent` instead of the `VirtualPlayerModel` that exists right now. Games should now be able to function without the player model at all because of the work I have recently done. Pretty much we would create the game like usual but we would not give a viewer component to the second player.

Theres some other small things I have questions about so I will make a review at some point.

Lunarmagpie:
I just took another look at this pull request! Starting off on a positive note, making `getTurnActions` a generator was really clever.

The main architectural issue I have is that sound effects feel like they are tacked on. Since I want to add sound effects for other actions such as placing a card for normal gameplay, I think its time to discuss the sound effect issue. It seems to me that we need some sort of way for the backend to add sound effects then the frontend can play them when it receives them.  I tried to handle all sound effects in the frontend for my sound effects PR that is opened right now but it really did not work. I think getting the sound effect issue squared away before coming back to this would be a good idea.

I need to take a closer look at the way the "boss landing" code works. I want to make sure it won't be that bad to modify if we add a second boss.

I think theres some minor code style issues but thats not really a big deal. I want @alicetouhou to review all the text before we merge this.

Lunarmagpie:
Closing in favor of #902

PR #475, opened 2024-03-10
Battle log fixes

Screaper91:
- Single use cards now show in battle log if they require a pick/modal request when played
- Armor Stand no longer causes crashes when attacked
- Minor log formatting fix when Armor Stand is placed


Benji42:
Is there a reason that each card with a modal/pick needs to add the battle log text within the card? Otherwise, it looks good to me

Screaper91:
I have moved all single use logging to `applySingleUse` because if the card was removed during `onApply` hooks (Piston/Fire Charge/Helsknight) it would silently fail. Single use cards with modal/pick or that attack can now pass text & format to provide additional detail to the log.

PR #482, opened 2024-03-23
Thorns sprites (reopen)

Rvtar:
I remade my fork so I could have a base branch, synced to main, to make my branches out of.

Same sprites as the old PR, the current sprites looked the same, thorn colors correspond to their token ranks.


Sense101:
I like this - however, I'm unsure if it would be better to have the sprites match their armor alternatives rather than the token cost, as that can change. So the order would then be iron, diamond, netherite. Any thoughts on this?

Rvtar:
I did realize that quickly after I finished the sprites so I decided to see if someone would say the same thing. I will update them to be iron-diamond-netherite.

Rvtar:
Should I also make the sprites 16×16?

Sense101:
Looks good - congrats on your first PR!

Sense101:
I just saw  this "How too make iron not resemble paper."
Are you happy with how it looks now?

Benji42:
i think it'd look pretty good for iron to just have the normal black spikes, it was the original thorns after all

Rvtar:
> i think it'd look pretty good for iron to just have the normal black spikes, it was the original thorns after all

The issue would be the resemblance to Netherite.

Rvtar:
> I just saw this "How too make iron not resemble paper." Are you happy with how it looks now?

I am.

PR #483, opened 2024-03-23
Pearl-swap fix

Rvtar:
Fix the issue of Pearl's ability being able to trigger despite flipping heads after swapping by using 2 status effects to track the state of whether heads was recently flipped and by whom. Differentiation is done by comparing player IDs. The icons are intended not to be shown, relying on the feature in Benji's PR, so I put my placeholder icons for those.


Benji42:
Are we sure this is a bug? I was under the impression that switching to a different pearl was a valid strategy-- this is prob something to bring up in #meta-balancing on the server.

Rvtar:
There are bug reports about this and no one said it isn't. Everyone I faced thought it is.

Benji42:
So it looks like people do think this is a reasonable change, however before I merge it there are a few changes I would like you to make.
- Please merge the current dev branch into this so you can make the visibility feature work. You also need to fix merge conflicts.
- The status effects are needlessly complicated. You should only need one. You can look at what I did for clock and copy that. Also make sure that you're only removing the hooks that you create.
- The same system should also be in place for Joe, as these cards should be consistent.

Rvtar:
Reduced to a single status effect and did some cleanups. Should be ready for merging.

Benji42:
I asked sense to review it, so hopefully he gets to it soon

Sense101:
The current implementation is already intended behaviour.

PR #491, opened 2024-04-03
afterAttack fix

Rvtar:
Make abilities using afterAttack work with Target Block. I hope I haven't missed any cards.


Benji42:
Oh, I'm working on a different solution for the target block issues. I'm gonna close this but I appreciate the effort.

PR #492, opened 2024-04-03
Improvement in weakness system and attack modal

Benji42:
I made some changes to the attack modal
- `createWeakness` is now a tri-lean property (i don't think that's a word but it should be). The values are 'always'/'ifWeak'/'never'. This allows it to work with normal attacks and weakness potion without needing to create a new attack. A lot of interactions are simplified a lot, and some broken ones finally work now.
- I added another property `redirected`. This becomes true if the attack is redirected with an effect, so it can work with chainmail properly. I don't think this is a perfect solution, but after considering all the options I think it's the best one. My solution before was very messy since it involved changing attack types.


Rvtar:
https://en.wikipedia.org/wiki/Three-valued_logic

I think this one fits best regarding the tri-lean thing.

PR #494, opened 2024-04-04
Gem knockback fix

Rvtar:
A piece of the afterAttack fix that also fixed Gem not deleting knockback.


Rvtar:
Added a fix for the same issue with Chorus Fruit. It now gets discarded but the second SU has to be used before you switch.

Screaper91:
Please make the following changes:

- [ ] Gem should remove `afterAttack` hook at end of `afterAttack` instead of `onTurnEnd` to prevent playing additional single use cards after subsequent single use attacks  _(E.g. Use Fishing Rod, attack with Geminislay, then use Bow, then use Sword...)_

- [ ] Chorus Fruit should call `applySingleUse` during `onAttack` so it runs before Gem discards the card

Rvtar:
> Please make the following changes:
> 
> * [ ]  Gem should remove `afterAttack` hook at end of `afterAttack` instead of `onTurnEnd` to prevent playing additional single use cards after subsequent single use attacks  _(E.g. Use Fishing Rod, attack with Geminislay, then use Bow, then use Sword...)_
> * [ ]  Chorus Fruit should call `applySingleUse` during `onAttack` so it runs before Gem discards the card

Done.

Screaper91 (approved):

Benji42:
Hey I'm just wondering why there's some changes to some files that shouldn't be related to this change? You may need to cherry pick your commits onto the most recent version of the dev branch.

Rvtar:
So... it's been a while. Should I make a new PR form anew branch?

Benji42:
this looks good to me now 👍

common/cards/default/hermits/geminitay-rare.ts
```diff
@@ -32,27 +32,32 @@ class GeminiTayRareHermitCard extends HermitCard {
...
 			if (attack.id !== this.getInstanceKey(instance) || attack.type !== 'secondary') return
 
-			// To keep this simple gem will discard the single use card, if it's used
-			if (player.board.singleUseCardUsed) {
+			player.hooks.afterAttack.add(instance, (attack) => {
+				// To keep this simple gem will discard the single use card, if it's used
+				// Hopefully, removing it doesn't break anything.
 				discardSingleUse(game, player)
-			}
+				player.board.singleUseCardUsed = false
```
Sense101, line 44: This is unnecessary, discardSingleUse does this

common/cards/default/single-use/chorus-fruit.ts
```diff
@@ -29,7 +26,15 @@ class ChorusFruitSingleUseCard extends SingleUseCard {
 				[`attack `, 'plain'],
 			])
 
-			player.hooks.afterAttack.remove(instance)
+			player.hooks.afterAttack.add(instance, (attack) => {
+				// Remove change active hermit from the blocked actions so it can be done once more
+				game.removeBlockedActions('game', 'CHANGE_ACTIVE_HERMIT')
```
Sense101, line 31: This needs to straight away remove the afterAttack hook so it can't get called again

common/cards/default/hermits/geminitay-rare.ts
Sense101, line 44: This is unnecessary, discardSingleUse does this

common/cards/default/hermits/geminitay-rare.ts
```diff
@@ -32,27 +32,32 @@ class GeminiTayRareHermitCard extends HermitCard {
 		const {player} = pos
 
 		// @TODO egg confusion, and how can we get rid of follow up
+		// The follow-up is now present for the counterpart.
 		// is that even in the scope of this refactor?
-		player.hooks.afterAttack.add(instance, (attack) => {
+		player.hooks.onAttack.add(instance, (attack) => {
```
Sense101, line 37: Why does this need to be changed for a hermit card?

Rvtar: So she wouldn't delete Knockback before it is applied but she would still delete it.

common/cards/default/hermits/geminitay-rare.ts
Rvtar: Got it.

common/cards/default/single-use/chorus-fruit.ts
Rvtar: I believe nothing in the game could trigger afterAttack for a second time when you're using Chorus Fruit. Also, if I just delete it in afterAttack, a damage effect would cancel Chorus Fruit.

Sense101: The afterAttack hook is called for every section of an attack, it's bad practice for e.g. "game.removeBlockedActions('game', 'CHANGE_ACTIVE_HERMIT')" isn't run more than once for a *single* attack loop, as it wouldn't achieve anything. Therefore the afterAttack hook should still self remove.

I agree with the change made here to allow it to work for multiple attack loops, but you should also add a check to see if you've already applied the card before applying it again.

common/cards/default/hermits/geminitay-rare.ts
Sense101: This won't work, completedActions are not yet set in onAttack and therefore can't yet be modified. I'm also still not sure exactly what you're fixing by changing to onAttack here

Screaper91: This does modify `completedActions` during `afterAttack`, the hook must be added here so it is executed after Knockback. Otherwise Knockback will 'apply' after Geminislay, since hermits are attached before single-use effects and Knockback shouldn't _need_ first priority in `afterAttack`.

common/cards/default/single-use/chorus-fruit.ts
Rvtar: The card is applied in onAttack, which gets deleted in onDetach when Gem deletes the card. Here, onAttack summons afterAttack which keeps the option to switch because attacks remove that option. If afterAttack removes itself, there will be nothing to reallow the option to switch, though we actually can still call it intentional because there had been no opinions about this unless I missed something in the javascript era.

Screaper91: I made a PR that fixes the issues @Sense101 is referring to here [https://github.com/Rvtar/hc-tcg/pull/2](url). Using chorus fruit with Anvil, Anvil Drop, or Fire/Poison causes attack hooks to run too often.

Screaper91 (changes requested):

common/cards/default/single-use/chorus-fruit.ts
Rvtar: Thanks, I was completely tunnel-visioned on Gem+Chorus Fruit.

Screaper91 (approved):

Rvtar:
Do I resolve the comments?

Screaper91:
I think so at this point, since Sense hasn't replied in this PR for a month

Sense101:
Taking a look now

Sense101 (changes requested):

common/cards/default/hermits/geminitay-rare.ts
```diff
@@ -31,26 +31,29 @@ class GeminiTayRareHermitCard extends HermitCard {
...
 		const {player} = pos
 
-		player.hooks.afterAttack.add(instance, (attack) => {
+		player.hooks.onAttack.add(instance, (attack) => {
 			if (attack.id !== this.getInstanceKey(instance) || attack.type !== 'secondary') return
 
-			// To keep this simple gem will discard the single use card, if it's used
-			if (player.board.singleUseCardUsed) {
+			player.hooks.afterAttack.add(instance, (attack) => {
+				// To keep this simple gem will discard the single use card, if it's used
```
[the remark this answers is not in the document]
Rvtar: It's for a case of Knockback where it would apply after that line runs. Also, either way there would be a caution in applying SUs in afterAttack.

Sense101, line 38: As this is no longer accurate, just remove the "if it's used", then it's good to go.

Sense101 (approved):

PR #496, opened 2024-04-10
Fixes health reducing after Ladder

Screaper91:
Adds `healRow` function to "common/utils/board.ts"
- Fixes where healing can reduce health after using Ladder
_Rare False uses Supremacy after Ladder and clamps health on flipping heads_
- Fixes hermit taking extra damage if first attack is less than the extra health from Ladder
_A 250hp max Hermit ladders to get 300hp, if the first attack executed deals less than 50dmg then health would be set to 250_


Sense101:
Apologies, I fixed this before I saw this PR. Thank you anyways for your contribution.

PR #504, opened 2024-04-19
Change looting back to it's S1 variant

Benji42:


Rvtar:
Shouldn't the pick request prompt be "steal" instead of "discard"?

Benji42:
yep, fixed

Sense101:
I thought we were going with opponent's active Hermit?

PR #505, opened 2024-04-20
Cards now specify if a slot can be moved to

Screaper91:
Fixes cards not being able to be moved by opponent after placement


Sense101:
I can't believe that you did this, but you thought of the pretty much the same idea we did - we're currently working on porting over all the cards to a new canAttach system on the canAttachRefactor branch. Please feel free to open a pr on that fixing some of the cards, so far Benji's done most of the SU cards and I'll be working on the ones with @ NOWTODO in them

PR #509, opened 2024-04-23
Cancel button for api-made matches

Tyrannicodin:
This adds a cancel button and a privateCancel hook to notify the bot when a game is cancelled.


Lunarmagpie:
I don't think it's a very good idea to harcdode to the bot url.
Instead set it up so the URL is provided when you create a match. So you'd send a packet like
```
{
     cancel-webhook: "https://bot/api/on_cancel"
```
when a new match is created.

Tyrannicodin:
I actually removed the use of botUrl in the config and moved it to apiKeys.json, I just forgot to remove from the config file.

PR #513, opened 2024-04-30
Update wolf to only deal damage when actually attacked

Benji42:


Sense101:
Not merging this in favour of using calculateDamage - sorry about the miscommunication

PR #517, opened 2024-05-02
Fix Scar reviving by Deathloop repeatedly

Screaper91:


Benji42:
Is this meant to be opened still? You can make a PR to my [Benji42:onHermitDeathFixes](https://github.com/Benji42/hc-tcg/tree/onHermitDeathFixes) branch instead if you want to fix scar there instead.

Screaper91:
This is from when the changes were being done directly on the dev branch. 
I made another pull here [https://github.com/Benji42/hc-tcg/pull/2](url) after Sense force pushed them back.

Benji42:
Ah thanks didn't see 👍 In the future if I don't review quickly let me know on discord, likely that I just missed it

PR #524, opened 2024-05-03
Remove the onHermitDeath Hook

Benji42:


common/cards/default/hermits/goodtimeswithscar-rare.ts
```diff
@@ -40,41 +40,34 @@ class GoodTimesWithScarRareHermitCard extends HermitCard {
 			const attackerInstance = attack.getAttacker()?.row.hermitCard.cardInstance
 			if (!attackerInstance) return
 			// If this instance is not blocked from reviving, make possible next turn
-			if (canRevives[attackerInstance] == undefined) canRevives[attackerInstance] = true
+			if (canRevives[attackerInstance] === undefined) canRevives[attackerInstance] = true
 		})
 
 		// Add before so health can be checked reliably
 		opponentPlayer.hooks.afterAttack.addBefore(instance, (attack) => {
+			console.log(canRevives)
```
Sense101, line 48: Please remove console log

Sense101 (changes requested):
I don't like the changes made here to the ailments - adding new code inside the checkHermitHealth method is a band-aid that has the same issues the onHermitDeath hook did - namely that other cards cannot interact with this. Also, I know it's more verbose to add an additional hook in every status effect, but it does help a lot in terms of cards being fully understandable a lot faster than with a bunch of flags

common/cards/default/hermits/goodtimeswithscar-rare.ts
```diff
@@ -97,13 +89,16 @@ class GoodTimesWithScarRareHermitCard extends HermitCard {
...
 				player.hooks.onTurnStart.remove(instance)
-				if (canCleanUp()) player.hooks.onHermitDeath.remove(instance)
+				if (canCleanUp()) player.hooks.afterAttack.remove(instance)
 			})
-			player.hooks.onHermitDeath.add(instance, (hermitPos) => {
-				const hermitInstance = hermitPos.card?.cardInstance
-				if (hermitInstance) delete canRevives[hermitInstance]
-
-				if (canCleanUp()) player.hooks.onHermitDeath.remove(instance)
+			player.hooks.afterAttack.add(instance, (attack) => {
```
Screaper91, line 94:
```suggestion
			player.hooks.afterDefence.add(instance, (attack) => {
```
Please add a similar hook in onAttach so targetInstance is deleted when your hermit dies from any cause and after Totem has a chance to revive the row

Screaper91 (changes requested):
The next best hook instead of `onHermitDeath` is `afterDefence` to handle when your hermits die. Alternatively, maybe have `onHermitDeath` called at the end of the attack loop to still allow XB/Golden Axe to restrict which hooks are called.

Should Scar and Totem also check if status effects are removed on death? Check if a player can use Clock consecutively if their hermit dies by Fire and is revived.

PR #525, opened 2024-05-07
Remove regex validation for deck titles

Benji42:
Not merging this myself, because there may be a problem with doing this I'm not aware of.

However, I can't find any and this feature is heavily requested.


client/src/app/deck/deck-edit.tsx
```diff
@@ -98,13 +98,9 @@ const DeckName = ({loadedDeck, setDeckName, isValid}: DeckNameT) => {
 				placeholder="Enter Deck Name..."
 				className={css.input}
 				required={true}
-				pattern={`^[a-zA-Z0-9 ]*$`}
 				onBlur={() => handleBlur()}
 				data-focused={inputIsFocused}
 			/>
-			<p className={css.errorMessage}>
-				Deck name should be between 1-32 characters and shouldn't include any special characters.
-			</p>
```
Lunarmagpie, line 107: Since this is stored server side, its a good idea to have a maximum length. Otherwise, somebody could crash the server by naming the deck something massive.

Sense101:
There are problems with this, it opens up the possibility of sql injection-like attacks

Lunarmagpie:
> There are problems with this, it opens up the possibility of sql injection-like attacks

I don't think its making the problem worse. Something like `DELETE TABLES` was totally doable before and your input should be properly sanitized anyway.

Sense101:
Would it not still be better to simply enforce characters, but allow more than before? Also yes to a character limit

Benji42:
I think if that's an issue I should close this PR and we should solve the root of the problem first. But, I think firebase doesn't let you inject commands in the first place but I would need to double check. We are using a library for it iirc and those are usually made to be safe. It should be a non issue (except the character limit)

Benji42:
Looked it up, firebase is designed so something similar to injection is impossible from the start. thought this was the case.  will fix the length issue then i'm going to merge this

PR #529, opened 2024-05-09
Improve Battle Log entries

Screaper91:
- Adds entries for reviving a hermit with Totem and "Deathloop".
- Allows hermit attack entries to be edited before being sent in chat.


Benji42:
Do you think we could move over coinflips to use `modifyHermitAttackDescription` as well? It would be more compact.

Screaper91:
I would create a new function to add the coin flips to an entry since we want the result to be shown after the animation plays.
Other scenarios to consider:
- "Sheep Stare" and "Watch This" may flip again after the first flip is heads
- "Trap Hole" flips for each single-use played (Piston, Fire Charge, and "Geminislay" allow multiple to played in one turn)
- "Aussie Ping", "Sheep Stare", and "Trap Hole" flip during the opponent's turn

Benji42:
in that case this looks good. i don't think we need to worry about modifying the coinflip system immediately.

Benji42 (commented):
I don't like how `modifyHermitAttackDescription` and `modifyHermitAttackTarget` work. We shouldn't have to mutate battle log entries after they're created.

Benji42:
I decided to rework this system myself since there's a lot to be changed.

however, you can take a shot too if you'd like

Benji42:
Superseded by #538

PR #537, opened 2024-05-22
Original weakness

Rvtar:
An implementation of weakness that matches how it works on Hermitcraft. The weakness icons will show up at hermits with types that have their weaknesses modified.

The feedback was kinda mixed so I might as well try.


Rvtar:
Just realized that onCardPlay is just a bootleg onAttach that gives the whole card instead of just cardInstance. Gonna leave it on for now because it makes this thing a whole lot simpler and changing onAttach to input that would probably be a refactor a that point.

Sense101:
Yeah we don't want to be adding unnecessary hooks, I'm sure you can find a way to do this with onAttach (e.g. is there a way you could use the card instance to get more information about the card?)

Benji42:
`getCardPos` is probably what you want.

Rvtar:
Yeah, I completely missed it the first time.

const card = gatCardAtPos(game, getBasicCardPos(game, instance)) should work.

Rvtar:
(done)

Rvtar:
1) I have prettier installed.
2) I formatted all of the listed documents.

What is going on?

Rvtar:
It's good now.

Sense101:
Why are you calling it weakness dummy? Can you please clarify what exactly is changed by moving over weakness to the original?

Rvtar:
There's a brain part that is invisible and the UI (dummy) that's visible. Cannot have both properties at the same time. I'm not sure of applying the brain to oblivion would cause any issues but if doable, it can be reduced to a single one.

Rvtar:
For the mechanics, the strengths and weaknesses are tied to the types themselves rather than the hermits.

Benji42:
This feels a bit overly-complicated to me. Why can't the weakness dummy stuff be on the weakness potion?

But to be honest, I think the real change that needs to be made here is to allow status effects to have multiple targets. It would allow this to be implemented a lot nicer, I think. I would be happy with it then.

common/status-effects/weaknessRef.ts
```diff
@@ -0,0 +1,148 @@
...
+					}
+				}
+			}
+		}
+
+		if (!statusEffectInfo.duration) statusEffectInfo.duration = this.duration
+
+		applyDummies()
+
+		STRENGTHS[strongType].push(weakType)
```
Lunarmagpie, line 68: This doesn't work. This modifies the weak types for everyone changing the game.
This variable is supposed to be immutable.

Lunarmagpie:
I suggest implementing this by changing the weakness of the incoming attacks  to the hermit to `always`.

Rvtar:
I realized I shouldn't have pushed this stuff rn.

Rvtar:
I'll open again when I'm done.

PR #538, opened 2024-05-22
Rework battle log and chat system

Benji42:
This PR reworks the battle log and chat system.
- Added a parser for parsing the chat, battle logs, and card descriptions. This is a lot neater than the mix of systems implemented before, has no bugs, and is also a lot more expandable.
- Updated battle log system.
  - Instead of an array, logs now use the parser so they're a lot more readable and easier to modify.
  - Now, each card has their own log that is sent when the card is placed in a non-SU slot, or applied.
  - Attacks also have their own logs.
  - There is a function to create custom logs for other situations.
- Ripped out old profanity filter and replaced it with a new one. It's a lot faster now and does not have the same issues.
- Fixed a lot of buggy behavior in chat log.

developed with @Lunarmagpie


common/cards/alter-egos/hermits/goatfather-rare.ts
```diff
@@ -36,20 +36,25 @@ class GoatfatherRareHermitCard extends HermitCard {
...
 
 		const {player, opponentPlayer, row, rowIndex} = pos
 
-		if (attacks[0].type !== 'secondary' || !row?.hermitCard) return attacks
+		if (attack.type !== 'secondary' || !row?.hermitCard) return
 
 		const coinFlip = flipCoin(player, row.hermitCard)
 
-		if (coinFlip[0] === 'tails') return attacks
+		if (coinFlip[0] === 'tails') return
```
Screaper91, line 49: Goatfather should return `attack` when using primary or Anvil Drop flips tails, this prevents those attacks from executing

Benji42: Hmm, fixed this but now i'm wondering. Would it be possible to make it so getAttacks can't return null? Would prevent this type of mistake from happening. There's probably an edge case I'm not thinking of, though

Benji42:
I opened the PR for review but that doesn't mean it's ready to merge yet. We still need to go through and make sure the logs are all correct, and confirm there's no bugs.

common/cards/alter-egos/hermits/goatfather-rare.ts
Screaper91: Maybe have `getAttacks` return `AttackModel | null` so it must explicitly return a value, even if it not an attack

common/types/cards.ts
```diff
@@ -60,3 +63,20 @@ export type SlotPos = {
...
+		name: string
+		hermitCard: string
+		slotType: string
+	}
+	pick: {
+		rowIndex: string
+		name: string
+		slotType: string
+	}
+}
```
Lunarmagpie: Sorry this was my bad. This type name is awful. I am not sure what a good type name is.
Also please document the fields.

common/utils/attacks.ts
```diff
@@ -147,77 +147,69 @@ function shouldIgnoreCard(attack: AttackModel, instance: string): boolean {
...
+
+	// STEP 1 - Call before attack and defence for all attacks
+	runBeforeAttackHooks(attacks)
+	runBeforeDefenceHooks(attacks)
+
+	// STEP 2 - Call on attack and defence for all attacks
+	runOnAttackHooks(attacks)
+	runOnDefenceHooks(attacks)
+
+	// STEP 3 - Execute all attacks
```
Lunarmagpie, line 164: These comments don't add clarity. The function names and forEach does all of this for us.
```suggestion
	runBeforeAttackHooks(attacks)
	runBeforeDefenceHooks(attacks)

	runOnAttackHooks(attacks)
	runOnDefenceHooks(attacks)

```

common/utils/attacks.ts
```diff
@@ -147,77 +147,69 @@ function shouldIgnoreCard(attack: AttackModel, instance: string): boolean {
...
-		runAfterDefenceHooks(allAttacks)
+		attack.nextAttacks = []
+	})
 
-		// STEP 7 - If we added any new attacks in afterAttack or afterDefense, loop around
-		for (let i = 0; i < allAttacks.length; i++) {
-			attacks.push(...allAttacks[i].nextAttacks)
-		}
+	if (!withoutBlockingActions) {
+		// STEP 5 - All attacks have been completed, mark actions appropriately
```
Lunarmagpie, line 180: I would keep this comment

common/utils/attacks.ts
```diff
@@ -147,77 +147,69 @@ function shouldIgnoreCard(attack: AttackModel, instance: string): boolean {
...
+			'PLAY_SINGLE_USE_CARD',
+			'CHANGE_ACTIVE_HERMIT'
+		)
 	}
+
+	// STEP 6 - After all attacks have been executed, call after attack and defence hooks
+	runAfterAttackHooks(attacks)
+	runAfterDefenceHooks(attacks)
+
+	// STEP 7 - Execute new attacks created in afterAttack hooks
```
Lunarmagpie, line 196: I also like this comment

tests/common/utils/formatting.test.ts
```diff
@@ -0,0 +1,63 @@
+import {describe, expect, test} from '@jest/globals'
```
Lunarmagpie, line 1: For me, add tests to make sure error handling works properly

Lunarmagpie (approved):

common/utils/attacks.ts
Benji42: Going to keep the comments bc they were like that beforehand. Unless you strongly feel they need to be changed

Lunarmagpie: If they were like that before, I think it's fine to leave them

Benji42:
This should be done now.

common/utils/attacks.ts
Sense101: I wrote those comments, and for me they add clarity. Generally I wouldn't remove other comments unless they are clearly outdated

common/cards/alter-egos-ii/hermits/iskallman-rare.ts
```diff
@@ -145,10 +147,17 @@ class IskallmanRareHermitCard extends HermitCard {
 			})
 			attack.addNewAttack(backlashAttack)
 
+			const attackerInfo = HERMIT_CARDS[attacker.row.hermitCard.cardId]
 			const hermitInfo = HERMIT_CARDS[pickedRow.hermitCard.cardId]
 			if (hermitInfo) {
 				const maxHealth = Math.max(pickedRow.health, hermitInfo.health)
 				pickedRow.health = Math.min(pickedRow.health + 50, maxHealth)
+				game.battleLog.addCustomEntry(
```
Sense101, line 155: As this is used in so many places now, maybe it's a good idea to rename it to simply "addEntry"?

common/cards/alter-egos/single-use/egg.ts
```diff
@@ -49,7 +51,7 @@ class EggSingleUseCard extends SingleUseCard {
 					if (!pickResult.card) return 'FAILURE_INVALID_SLOT'
 
 					// Store the row index to use later
-					player.custom[targetKey] = rowIndex
+					player.custom[targetKey] = pickResult
```
Sense101, line 54: What does this change? Looks like it'd probably fix a targeting bug, if yes, what is it as we'd want to add it in the next changelog

common/cards/alter-egos/single-use/ender-pearl.ts
```diff
@@ -48,25 +50,27 @@ class EnderPearlSingleUseCard extends SingleUseCard {
 				// We need to have no card there
 				if (pickResult.card) return 'FAILURE_INVALID_SLOT'
 
+				const activeRow = getActiveRowPos(player)
+				if (player.board.activeRow === null || !activeRow) return 'FAILURE_INVALID_DATA'
+
+				const logInfo = pickResult
```
Sense101, line 56: What type is logInfo, and what's it used for? Unclear to me

common/cards/alter-egos/single-use/ladder.ts
```diff
@@ -83,7 +83,7 @@ class LadderSingleUseCard extends SingleUseCard {
 				}
 
 				// Apply
-				applySingleUse(game, [])
+				applySingleUse(game)
```
Sense101, line 86: I like that we don't have an unneeded parameter here anymore

common/cards/base/card.ts
```diff
@@ -27,12 +28,16 @@ abstract class Card {
 	public name: string
 	public rarity: CardRarityT
 
+	/** The battle log attached to this attack */
```
Sense101, line 31: This should say attached to this *card*. Also could I get clarification, ideally in a comment here, when this is used and what the value should be set as

common/config/debug-config.example.json
```diff
@@ -14,5 +14,6 @@
 		"enabled": false,
 		"clearConsole": true
 	},
-	"autoEndTurn": false
+	"autoEndTurn": false,
+	"logAttackHistory": false
```
Sense101, line 18: Thanks for adding this

common/utils/board.ts
```diff
@@ -154,12 +151,14 @@ export function applySingleUse(
 
 	currentPlayer.board.singleUseCardUsed = true
 
-	game.battleLog.addApplyEffectEntry(effectAction)
 	currentPlayer.hooks.onApply.call()
 
 	// This can only be done once per turn
 	game.addCompletedActions('PLAY_SINGLE_USE_CARD')
 
+	// Send the logs
```
Sense101, line 159: Comment is send the logs but this isn't sending the logs, is it?

Sense101 (commented):
I like that you went through and removed all the /n/n.
I also kinda like that getAttacks now only returns one attack, although can you explain why this change was made please? In addition, now that this was done, it might be better to rename the function on the cards to getAttack

Lunarmagpie:
> I also kinda like that getAttacks now only returns one attack, although can you explain why this change was made please? In addition, now that this was done, it might be better to rename the function on the cards to getAttack

The old system had a problem where you could either have multiple attacks by returning an array or child attack. Having multiple ways to do things is bad, so the array way was removed as it was making algorithms very difficult to write.

common/cards/alter-egos-ii/hermits/iskallman-rare.ts
Benji42: Sure

common/cards/alter-egos/single-use/egg.ts
Benji42: I'm not sure. This just made egg easier to program with the new log changes. Egg was written a bit strangely before. Would appreciate if you could send me a link to the bug report so I can confirm it would fix it.

common/cards/alter-egos/single-use/ender-pearl.ts
Benji42: Yeah.... this is the only part of the update I don't like.

I'm passing in pickInfo into applySingleUse when it exists so we can use it in the descriptions of cards. However, some cards make you pick a slot and we need to put the card you picked instead so I'm creating a new variable to override the card before passing it into applySingleUse. This is kind of a temporary solution, I don't mind removing it for now and coming back to those logs later.

common/utils/board.ts
Benji42: Yeah, typo

Benji42:
> The old system had a problem where you could either have multiple attacks by returning an array or child attack. Having multiple ways to do things is bad, so the array way was removed as it was making algorithms very difficult to write.

To expand on this, It made programming some of the logging stuff a bit complicated because we needed to handle multiple ways of adding attacks. Now there's just one consistent way so it's a lot simpler and less prone to error

common/cards/base/card.ts
Benji42: 👍

client/src/components/formatting/formatting.tsx
```diff
@@ -0,0 +1,82 @@
...
+				{settings.isOpponent
+					? nodeToHtml(differentTextNode.opponentText, settings)
+					: nodeToHtml(differentTextNode.playerText, settings)}
+			</span>
+		)
+	} else if (node.TYPE == 'ProfanityNode') {
+		let profanityNode = node as ProfanityNode
+
+		if (settings.censorProfanity) {
+			return <span> {profanityNode.censor()} </span>
```
Screaper91, line 58: This causes errors since `profanityNode.censor()` cannot be serialized.

common/models/battle-log-model.ts
```diff
@@ -1,321 +1,329 @@
...
+	public async sendLogs() {
+		while (this.logMessageQueue.length > 0) {
+			const firstEntry = this.logMessageQueue.shift()
+			if (!firstEntry) continue
 
 			this.game.chat.push({
 				createdAt: Date.now(),
-				message: lastEntry.description,
-				playerId: lastEntry.player,
+				message: formatText(firstEntry.description),
```
Screaper91, line 77: Any player names in logs are not tested for profanity
```suggestion
				message: formatText(firstEntry.description, {censor: true}),
```

server/src/routines/background/chat.ts
```diff
@@ -23,55 +16,17 @@ function* chatMessageSaga(game: GameModel, action: AnyAction) {
...
-		return {
-			text: chunk.message,
-			censoredText: profanityFilter(chunk.message),
-			format: 'plain',
-		}
-	})
-
 	game.chat.push({
+		message: concatFormattedTextNodes(
+			formatText(`$p${game.players[playerId].name}$ `),
```
Screaper91, line 21: Player names are not tested for profanity
```suggestion
			formatText(`$p${game.players[playerId].name}$ `, {censor: true}),
```

Screaper91 (changes requested):
There are errors relating to profanity censoring, suggested fixes at [https://github.com/Benji42/hc-tcg/pull/5](url)

Benji42:
@Screaper91 Oh... had no CI so I didn't realize, but your PR causes all the unit tests to fail. I'm going to unmerge it and you can re open it.

server/src/routines/background/chat.ts
Lunarmagpie: I agree with adding this! I left it out before but it does make sense considering past history!

Benji42: Yeah. Just make sure you are also censoring player names in all the other places they are displayed if you make this change.

Lunarmagpie:
Im gonna pull in that PR actually and make the fixes manually. Its a good start so I want to work off of it.

Lunarmagpie:
I merged the PR in

common/cards/alter-egos/single-use/egg.ts
Sense101: I'm not aware of a bug report, but this needs testing to clarify what it might change

common/cards/alter-egos/single-use/egg.ts
```diff
@@ -63,15 +65,12 @@ class EggSingleUseCard extends SingleUseCard {
 			const activePos = getActiveRowPos(player)
 			if (!activePos) return []
 
-			const targetIndex: number = player.custom[targetKey]
-			if (targetIndex === null || targetIndex === undefined) return
-			const targetRow = opponentPlayer.board.rows[targetIndex]
-			if (!targetRow || !targetRow.hermitCard) return
+			const pickInfo: PickInfo = player.custom[targetKey]
+			if (!pickInfo || !pickInfo.rowIndex) return
```
Screaper91, line 69:
```suggestion
			if (!pickInfo || pickInfo.rowIndex === null) return
```

common/status-effects/sleeping.ts
```diff
@@ -20,16 +20,21 @@ class SleepingStatusEffect extends StatusEffect {
 	}
 
 	override onApply(game: GameModel, statusEffectInfo: StatusEffectT, pos: CardPosModel) {
-		const {player, card, row} = pos
+		const {player, card, row, rowIndex} = pos
 
-		if (!card || !row?.hermitCard) return
+		if (!card || !row?.hermitCard || !rowIndex) return
```
Screaper91, line 25:
```suggestion
		if (!card || !row?.hermitCard || rowIndex === null) return
```

Screaper91 (changes requested):
`rowIndex` may be `0` and may cause issues when checking for `null`

common/cards/alter-egos/single-use/egg.ts
Benji42: It doesn't change anything that I'm aware of (besides for the bug that screaper found).

Benji42:
This is done unless anyone has any further suggestions/changes.

Lunarmagpie:
LGTM

PR #548, opened 2024-05-26
Untrack debug config

Benji42:
untracks debug config. this should hopefully fix some of the issues we're having with people accidentally pushing it


Sense101 (changes requested):
Glad you've finally managed to do this - however I think it should be called .prod.json instead of .example.json, as it's used for production deploys. Also, will this not mean if someone clones the code and does not create a local debug config, that it won't build/run?

Benji42:
1. no, example.json is standard, many other projects use it.
2. it's in the readme. shouldn't be hard to figure out. unfortunately we can't use require() for this so it's not possible to fix, i tried, so this is the best solution to the issue.

Sense101:
Alright, how about this: create a npm task to copy the debug config from the example config, and have the docker flow run that on deploy before building. Then you simply have to run that task when cloning fresh, instead of manually copying it

Benji42:
ah, i tried but it only works on linux wsl. is it possible to give a command that runs something different depending on operating system? otherwise i'm not sure that's worth it

Sense101:
I think there's a way to check what os is running but I think at that point it's not worth the hassle. Good to go :+1:

Sense101 (approved):

PR #549, opened 2024-05-26
Allow playing String and an item card

Screaper91:
Placing String in an opponent's item slot can now be done multiple times in the same turn and no longer prevents/is prevented by placing an item card on your side


Benji42:
I'm not sure this seems like the right solution. I think it would be better to rework `play-card.ts` to not have the issue in the first place.

Lunarmagpie:
Hello, I support Benji on this. The TODO statement makes me doubt the stability of this solution. I suggest investigating how to do this in a more stable/less bug prone way. The code needs people to investigate and fix old systems more than anything right now!

PR #561, opened 2024-06-09
Create Docker Build CI

Lunarmagpie:
Adds CI to ensure that the docker image builds.
This will slow down CI a bit, but I think it worth running at least before merging in the main branch.

### DIFF
--- a/.github/workflows/workflow.yaml
+++ b/.github/workflows/workflow.yaml
@@ -64,3 +64,14 @@ jobs:
 
       - name: Build
         run: npm run build
+
+  build-docker:
+    name: Test build docker image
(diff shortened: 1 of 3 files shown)


Benji42:
What is this supposed to do?

Benji42:
Yeah makes sense. I think this is a good addition. It's pretty bad we have docker issues that can delay releases after we say they're going to happen.

Lunarmagpie:
We may not need this, the main change that has to be made is either:
- Do our test build CI with the `NODE_ENV=production`
- Or build the docker image without `NODE_ENV=production`

Benji42:
yeah, sounds reasonable, LGTM

PR #562, opened 2024-06-09
Change Grian rare to using the selectCards modal instead of a custom modal

Benji42:
Hoping this improves the stability of the card. Could not replicate any of the bugs in testing.

In addition, I also centered the selectCards modal card as it looks a bit better.


Lunarmagpie:
LGTM

PR #564, opened 2024-06-09
Fix for Card Rendering Speed

Lunarmagpie:
This PR has a possible for the for the slow card render speed in the deck editor.

## Changes
- Remove animations for the left-hand-side cards. These animations made the system a bit slower and more importantly, made it feels really slow.
- The tooltip class was very unoptimized. I removed all the clones and made it so it only creates a floating window when necessary.
- All card components now use `memo`. Many identical cards are being rendered so this helps speed up the game a lot.

I believe these changes will also reduce lag in-game.


client/src/components/card-list/card-list.tsx
```diff
@@ -30,21 +31,20 @@ const CardList = (props: CardListProps) => {
...
-				}}
-			>
+			  }
+			: {}
+
+		return (
+			<CSSTransition timeout={250} unmountOnExit={true} classNames={cssClasses}>
 				<CardComponent
-					key={card.cardInstance}
+					key={card.cardId}
```
Tyrannicodin, line 47: I believe this would cause an error when there a multiple of the same card in one list.

client/src/components/card/card.tsx
Tyrannicodin, line 1: I don't know if I'm missing something here, but the changes to this file seem unnecessary.

Tyrannicodin (commented):
Looks mostly good, couple small things to double check.

client/src/components/card-list/card-list.tsx
Lunarmagpie: Fixed!

client/src/components/card/card.tsx
Lunarmagpie: Yeah, that was my bad

Tyrannicodin (approved):
Looks good, just need to run prettier.

Benji42:
Is this completely done now?

Lunarmagpie:
Yes

PR #570, opened 2024-06-10
Add CI To Release With Github Releases and Docker

Lunarmagpie:
This should make releases harder to mess up, theoretically. Tags will allow us to only have one main branch, this means PR could never be merged into the wrong place. I've used this system on a few projects in the past and it makes releases very easy.

Requires EVNs to be set up and there to be a dockerhub.


Benji42:
Hopefully we can set up automatic releases after this

Benji42 (approved):

PR #571, opened 2024-06-10
Add eslint

Tyrannicodin:


Lunarmagpie:
Now that we have biome, we will not be using eslint so I am closing this

PR #576, opened 2024-06-13
Change how armor stand uses discard

Benji42:
This is a huge issue in the code right now. Armor stand discarding itself allows for a `RowStateWithHermit` to NOT have a Hermit. With this change, it always will.

Now, only Hermit cards give a point when knocked out to compensate. I think this is fine. If we need a different system, we can discuss that in the future

I also simplified the attack loop, so attacks are NOT ran if they're added after the main attack loop is over. Thorns ect. uses `executeExtraAttacks` instead now.

### DIFF
--- a/server/src/routines/game.ts
+++ b/server/src/routines/game.ts
@@ -190,21 +190,27 @@ function* checkHermitHealth(game: GameModel) {
 				// Add battle log entry
 				game.battleLog.addDeathEntry(playerState, row)
 
-				if (row.hermitCard) discardCard(game, row.hermitCard)
-				if (row.effectCard) discardCard(game, row.effectCard)
+				const cardType = CARDS[row.hermitCard.cardId].type
+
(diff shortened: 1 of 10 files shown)


server/src/routines/game.ts
```diff
@@ -190,21 +190,27 @@ function* checkHermitHealth(game: GameModel) {
...
-				// reward card
-				const opponentState = playerStates.find((s) => s.id !== playerState.id)
-				if (!opponentState) continue
-				const rewardCard = playerState.pile.shift()
-				if (rewardCard) opponentState.hand.push(rewardCard)
+				// Only hermit cards give points
+				if (cardType === 'hermit') {
+					playerState.lives -= 1
+
+					// reward card
```
Lunarmagpie, line 208:
```suggestion
					// reword card
```

Lunarmagpie:
Visually looks good

PR #580, opened 2024-06-16
Fix piston not discarding itself

Benji42:
A little confused why this stopped working, did not change anything related to the card. I restructured it to be more similar to fire charge. Not sure if this would cause any issues.

### DIFF
--- a/common/cards/alter-egos/single-use/piston.ts
+++ b/common/cards/alter-egos/single-use/piston.ts
@@ -129,16 +129,6 @@ class PistonSingleUseCard extends SingleUseCard {
 				// Move the item
 				swapSlots(game, itemPos, targetPos)
 
-				// Only add the after apply hook here
-				player.hooks.afterApply.add(instance, () => {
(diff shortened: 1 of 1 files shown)


common/cards/alter-egos/single-use/piston.ts
```diff
@@ -153,6 +143,22 @@ class PistonSingleUseCard extends SingleUseCard {
 				delete player.custom[itemIndexKey]
 			},
 		})
+
+		player.hooks.afterApply.add(instance, () => {
+			console.log('afterApply')
```
Lunarmagpie, line 148: Remove this
(We really need to set up CI for this)

Lunarmagpie (changes requested):

Lunarmagpie (approved):
I didn't run the code to check, but it lgtm

Benji42:
yeah seems to work fine to me as well but there may have been a reason it was not implemented like this to begin with so i'm hesitant to merge it.

PR #581, opened 2024-06-16
Switch to Combinatory Logic based System For Setting Selectable Slots

Lunarmagpie:
This PR reworks card attach expressions and pick requests to use combinatory logic, this allows us to check what slots a card is placed and highlight the placeable cards on the frontend. In addition, the code is far easier to read and removes a lot of code.

Testing Progress (completed): https://docs.google.com/spreadsheets/d/1rEPBuN3ZP05Ahx771z0VghB_bkyZmf_Ek1V0hFiWSOk/edit?gid=1878436283#gid=1878436283 

Bug prevention:
- After this update, pick requests and attach conditions use the same exact syntax, which means we can put the same exact expression into both. This means it's now very easy to only make a card playable when the pick condition can be fulfilled. 

Bug fixes:
- Fixed a bug where string, when played in an item slot, could not be placed after you placed an item that turn, and would also not allow you to play another item after it
- Redid the available actions for playing cards system slightly, making the messages when they're available at the end of the turn much more intuitive
- Fixed a bug where composter would sometimes only make you discard one card
- Fixed a bug where composter would delete your card if cancelled after the first card was selected
- Fixed a bug where changing your active Hermit would not work properly after ladder was used
- Fixed a bug where Cubfan rare could not switch after efficiency was used
- Fixed a few unintuitive command block interactions with other cards
- Fixed a bug where armor stand could not be healed

Temporary changes:
- After this change, the card healing function references armor stand. This is NOT permanent. As this is only the first phase of the card rework, not all the systems are in yet. Our next change will focus on giving armor stand HP so it can be healed in a proper way, as well as redoing the healing system.


Benji42:
I think this was supposed to be a draft PR

Lunarmagpie:
There systems written for this PR are complete. There is no other expected changes.
QA is still ongoing but that should not really effect the review.

Benji42:
> There systems written for this PR are complete
open to change, though, there is definitely some bits that can be improved
particularly:
how available/blocked/completed actions work - It doesn't play that great with this, but I don't know if that's a problem with this PR or due to those systems not being great
Also, I think unless you understand FP well the learning curve is a bit higher than the old method. It's worth it because this is so much more powerful, but docs can possibly be improved
I don't like the `slots.locked` function and it's hook that much, but I'm at least confident in saying it's superior to the old method. If anyone has suggestions on how to improve it they are welcome

client/src/app/game/actions/actions.tsx
```diff
@@ -121,7 +122,10 @@ const Actions = ({onClick, localGameState, mobile, id}: Props) => {
 			dispatch(attackAction())
 		}
 		function handleEndTurn() {
-			if (availableActions.length === 1 || settings.confirmationDialogs === 'off') {
+			/** @TODO Rewrite everything, there is no world in which this should be compared to a constant.
```
Lunarmagpie, line 125: Remove this comment

PR #586, opened 2024-06-18
More connection fixes

Benji42:
I found a few more issues with the connection code, I'm hoping these changes fix the remaining issues.


Lunarmagpie (approved):
This should be correct

PR #588, opened 2024-06-19
Play sound on game start for both players

Lunarmagpie:
Fixes a bug where only player one would hear a sound on game start


Benji42:
double equals instead of ===

Lunarmagpie:
> double equals instead of ===

fixed

Benji42:
This might be part of the issue but I'm also pretty sure that there's other issues with this

Benji42:
Ok nevermind, seems to work consistently now

Benji42:
Jamm verified, I will merge this

PR #591, opened 2024-06-27
Make axe ignore for the whole turn again

Rvtar:
beforeAttack now doesn't get deleted when attacking.

### DIFF
--- a/common/cards/default/single-use/golden-axe.ts
+++ b/common/cards/default/single-use/golden-axe.ts
@@ -18,7 +18,8 @@ class GoldenAxeSingleUseCard extends SingleUseCard {
 	}
 
 	override onAttach(game: GameModel, instance: string, pos: CardPosModel) {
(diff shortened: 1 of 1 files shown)


common/cards/default/single-use/golden-axe.ts
```diff
@@ -41,12 +43,16 @@ class GoldenAxeSingleUseCard extends SingleUseCard {
...
+
+			attacking = true
+
 			if (!opponentActivePos) return null
 
 			if (attack.id === attackId) {
 				applySingleUse(game)
 			}
 
+			player.hooks.onAttack.add
```
Lunarmagpie, line 55: Please remove this line

Lunarmagpie (changes requested):

Lunarmagpie:
I tested the code and it looks good to me. (Made sure to test the issue when you can add and remove golden axe several times and get like +1000 damage)

PR #596, opened 2024-07-02
Rework cards to use composition based system

Lunarmagpie:
This PR reworks cards to use a composition based system. This removes hacky implementations/buggy behavior with Armor Stand, Milk Bucket, and Water Bucket. Additionally, this system should make any crazy ideas easier to add in the future.

Changes made:
- Removed card subclasses. These did not play well with cards that were between different classes. The new system uses a dictionary of properties that is different for each card type.
- Remove arrays `HERMIT_CARDS`, `EFFECT_CARDS`, etc. These arrays caused many bugs historically. You can use `CARDS` now but instances now have their properties attached to them which makes it much less important.  
	- Although we could have removed all arrays besides for `CARDS` without instead attaching props to card,  attaching props to cards also gets rid of needing to deal with undefined & the type system. It also reduces the total code footprint. Indexing arrays is just not very fun I suppose.
-  Added a `LocalCardInstance` class that allows card instances to be referenced on the frontend without worry that functions will be used that became undefined. Although the `WithoutFunctions` type is a bit strange, I think that its clear what the code does without understanding how it does it.  ]
- Errors are now thrown instead of ignored if a card is placed in an unsupported slot.
- Remove player.custom. In cases where values need to be shared between instances, the new `InstancedValue` class can be used. It does the  same thing but has typing. I found it was easier to follow as well. Only Rendog and Zombiecleo rare use this. I would like to remove this in the future but its out of scope for this PR. 
- Status effects use a new system that is more similar to cards.
- Move some hermit abilities to status effects. (Scar's can not revive effect, Zedaph's Sheep Stare, Pearl's Aussie Ping)
- Add typing to modals (by @Screaper91)


Screaper91:
Proposed changes to modal requests to fix grey screen caused by sending `CardInstance` instead of `LocalCardInstance`:
[https://github.com/Lunarmagpie/hc-tcg/pull/1](url)

Lunarmagpie:
> [https://github.com/Lunarmagpie/hc-tcg/pull/1](url)

Thank you for your work on this! I really like how you implemented the type system for modals. The union is very clever and I am a sucker for namespaces lol. I'm happy merging this into this branch as the entire thing will need to be tested anyway.

client/public/images/status/aussie-ping-immune.png
Lunarmagpie, line 1: Remove this

Lunarmagpie:
I would like to prevent multiple card instances for the same card from being allocated before merging this.

Lunarmagpie:
> I would like to prevent multiple card instances for the same card from being allocated before merging this.

After more investigation, this feels impossible without large changes that I feel are out of scope.

Lunarmagpie:
I feel confident about the code changes in this PR. We just need to test for bugs.

PR #597, opened 2024-07-06
Switch health card display to not be a card

Lunarmagpie:
Although, they share CSS, it did not make sense in the source code that health required a card to be displayed.


Benji42:
This looks good, but just make sure the tests are passed

Sense101:
Always wanted to do this, never got round to it. Lovely.

PR #602, opened 2024-07-11
Create status effect types and log fixes

Benji42:
- Status effects now have 4 types `normal`, `damage`, `system`, and `hidden`. This affect some of their properties.
- Logs also work properly now. Before, apply logs needed to be specified manually, and remove logs showed for all logs. Now the system is the same as the system cards and attacks use, so it's a lot easier to follow and works properly. Status effect logs can also be null for those that shouldn't make logs.


Lunarmagpie:
The code looks but i think `hidden` should be instead called `hiddenSystem` because it inherits the properties of `system`

Lunarmagpie:
LGTM

PR #603, opened 2024-07-12
Fix stacking damage effects

Screaper91:
Damage effects like "fire" and "poison" can no longer stack on the same hermit, and can again be added to a hermit with other non-damage effects


Lunarmagpie:
Thank you for the patch!

PR #608, opened 2024-07-18
Fix deck filtering and imporve performance

Lunarmagpie:
Improves deck filtering and also improves performance issues (theoretically). My computer is not happy but Im pretty sure this is more performant than the old code.


Benji42:
Please make sure Dream is disabled

PR #630, opened 2024-07-20
Status effect icons and log fixes

Benji42:
Updates the icons to Chimera Dev's icons and fixes some issues with a few logs.


DevChimera:
@Benji42 The current Invisibility potion heads seems to be the bad omen image.

PR #633, opened 2024-07-20
Make sure Hermit cards all return correctly in attack hooks

Benji42:


common/cards/default/hermits/cubfan135-rare.ts
```diff
@@ -33,7 +33,7 @@ class Cubfan135Rare extends Card {
 		const {player} = component
 
 		observer.subscribe(player.hooks.afterAttack, (attack) => {
-			if (!attack.isTargetting(component) || attack.type !== 'secondary') return
```
Lunarmagpie, line 36: THese are different

Benji42: was wrong before

PR #637, opened 2024-07-20
Single use fixes

Screaper91:
Fixes queries for the card in the single use slot to use `query.slot` and issues with playing a second single use in a turn


Lunarmagpie:
Thank you for the patch! People were having issues with single use cards that I couldn't reproduce, I think this probably fixes them!

PR #650, opened 2024-07-22
Update packages

Tyrannicodin:


Lunarmagpie:
Why does this add Terser?

Tyrannicodin:
Previously, terser was a required dependency. However now it is not but we still use parts that do require terser, therefore we must add it manually.

PR #657, opened 2024-07-22
Fix untyped cards not showing when there was an active type query in the deck builder

Benji42:


Lunarmagpie:
Yeah this is a good idea. It will help with wildcard.

Benji42:
Oh.. actually the way I implemented it doesn't show wild card. We'd need to change around how types work a bit for that to work. We can change it, but it'd be nice to do this after ae3 is merged then

PR #664, opened 2024-07-22
Fix Egg applying multiple times

Screaper91:
Fix egg repeating with multi-attack hermits and/or type advantage


Lunarmagpie (approved):

Lunarmagpie:
Thank you!

PR #665, opened 2024-07-22
Fix Deck Out

Lunarmagpie:
Fix deck out loss condition and improve display when the hand is empty. This was really dumb.


Screaper91 (changes requested):
"Get Good" doesn't trigger game end if it fails to draw a card, proposed fix: [https://github.com/Lunarmagpie/hc-tcg/pull/2](url)

Lunarmagpie:
Thanks for pointing that out. I did not notice that interaction. I don't think the patch you provided is the best way to go about fixing this, so I am going to search for a better solution.

PR #684, opened 2024-07-23
Disable all season X hermits

Benji42:
Not sure how someone got this in a game, so they're just disabled now to prevent that from happening at all


Lunarmagpie:
Good choice for now, we should re-investiagate once the production release is stable.

PR #690, opened 2024-07-23
Improve Clarity Of Attack Modal

Lunarmagpie:
This PR adds an extra button to the single use modal and attempts to show that attacking will also show the single use card. This also fixes visual uglyness when using a single use card without having any attacks available.


Benji42:
I want to mess around with colors a bit more before I merge this

Benji42:
Should be good to merge

PR #710, opened 2024-07-24
Switch to Preact

Lunarmagpie:
Surprisingly this just works? 
The performance seems to be a lot better, just judging from how much memory and CPU my browser uses.


Lunarmagpie:
Preact seems to render SVGs at a very low quality so I am closing this.

PR #713, opened 2024-07-24
Make Buttons Do Stuff

Lunarmagpie:
Adds keybinds to the game
- A = Attack
- E = End Turn
- C = Chat
- H = Show/Hide tooltips
- M = Mute

### DIFF
--- a/client/src/app/game/chat/chat.tsx
+++ b/client/src/app/game/chat/chat.tsx
@@ -1,4 +1,4 @@
-import {SyntheticEvent, useState} from 'react'
+import {SyntheticEvent, useEffect, useState} from 'react'
 import {useDispatch, useSelector} from 'react-redux'
 import {getChatMessages, getOpponentName} from 'logic/game/game-selectors'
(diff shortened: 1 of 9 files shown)


Benji42 (changes requested):
I think the button for tooltips should be T

client/src/app/game/actions/actions.module.scss
```diff
@@ -15,7 +15,7 @@
 	text-align: center;
 	padding: 15px 20px;
 	gap: 10px;
-	margin-top: 30px;
+	/* margin-top: 30px; */
```
Benji42, line 18: this doesn't really need to be commented out, you can just remove it

Benji42:
Before this is merged, you should probably add that it changes how muting works in the PR description. I can see us having to come back to this later if an issue happens with it

Lunarmagpie:
> I think the button for tooltips should be T

Done

Benji42 (approved):

PR #714, opened 2024-07-24
Fix Amnesia implementation

Screaper91:
- Fixes GA prize cards disabling user's attacks
- Fixes Amnesia disabling nothing when used by Ren/Cleo
- Fixes Amnesia blocking the same attack consecutively

### DIFF
--- a/common/cards/alter-egos-iii/hermits/architectfalse-rare.ts
+++ b/common/cards/alter-egos-iii/hermits/architectfalse-rare.ts
@@ -18,6 +18,7 @@ import {Hermit} from '../../base/types'
 type AttackInfo = {
 	attacker: CardComponent
 	attackType: HermitAttackType
+	turn: number
 }
 
 class ArchitectFalseRare extends Card {
@@ -65,6 +66,7 @@ class ArchitectFalseRare extends Card {
 				this.lastAttackInfo.get(game)[player.entity] = {
 					attacker: attack.attacker,
 					attackType: attack.type,
+					turn: game.state.turn.turnNumber,
 				}
 			})
 		)
@@ -82,7 +84,7 @@ class ArchitectFalseRare extends Card {
 				return
 
 			const lastAttackInfo = this.lastAttackInfo.get(game)[attack.target.playerId]
-			if (!lastAttackInfo) return
+			if (!lastAttackInfo || lastAttackInfo.turn !== game.state.turn.turnNumber - 1) return
 
 			if (lastAttackInfo.attackType === 'primary') {
 				game.components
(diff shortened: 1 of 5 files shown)


common/models/game-model.ts
```diff
@@ -11,6 +11,32 @@ import ComponentTable from '../types/ecs'
...
+	default: () => T
+	values: Record<string, T> = {}
+
+	public constructor(defaultFactory: () => T) {
+		this.default = defaultFactory
+	}
+
+	public set(game: GameModel, value: T) {
+		this.values[game.id] = value
+		game.task
```
Lunarmagpie, line 25:
```suggestion

```

Lunarmagpie (approved):
Thank you! I like how this is implemented.
I would consider combining InstancedValue and GameValue into one class. The type used as the key could simply be a generic. I would call this  base class `DefaultDict` because this is what its called in Python. if you don't want to make this change, or feel like its not a part of your patch, I'm happy to mark and something to do in the future.

common/cards/alter-egos-iii/hermits/architectfalse-rare.ts
```diff
@@ -37,36 +49,51 @@ class ArchitectFalseRare extends Card {
...
 
-	lastAttacker = new InstancedValue<CardComponent | null>(() => null)
-	lastAttackType = new InstancedValue<HermitAttackType | null>(() => null)
+	lastAttackInfo = new GameValue<Record<PlayerEntity, AttackInfo | undefined>>(() => {
+		return {}
+	})
 
 	override onCreate(game: GameModel, component: CardComponent) {
-		const {opponentPlayer} = component
+		if (Object.hasOwn(this.lastAttackInfo, game.id)) return
```
Lunarmagpie, line 57: Correct me if im wrong, but I believe this will always be false?

Screaper91: Most decks will have multiple copies of the same hermit, this should reduce allocating duplicate resources and prevent data loss if another GA CardComponent is created in the middle of the game

Lunarmagpie: I believe `Object.hasOwn` is checking the GameValue object and not the instance at the moment. i think for that intention this line needs to be `if (Object.hasOwn(this.lastAttackInfo.values, game.id))`.

Lunarmagpie:
I just noticed that items are not removed from the `lastAttackInfo` dictionary. I think you will need to add a `onDestroy` method for cards when the component is destroyed, then clear the item from the `lastAttackInfo` dictionary there.

Screaper91:
Currently components are never explicitly destroyed, only garbage collected when a game ends. That would require more work on the ecs system, but for now it will take a while for two games to generate the same `game.internalId`.

Lunarmagpie:
Thats true. Although it probably won't cause an issue, I don't like leaving memory around forever, and I also like taking advantage of garbage collection when we can. I think the best way to fix this would be created an 'onGameEndHook` we can subscribe to.

Lunarmagpie:
Thank you!

Lunarmagpie:
I just tested this and it works! I am ready to merge this.

PR #722, opened 2024-07-25
Fix Chainmail blocking Hermit specials

Screaper91:
- Fixes Chainmail negating damage from "Hyperspace", "Sheep Stare", and "Betrayed"


Lunarmagpie (approved):
Thank you! I didn't catch this.

PR #732, opened 2024-07-26
Improve Implementations of Both Skizzes

Lunarmagpie:
The previous implementations were buggy or poorly written. This is my attempt to redo them.


common/cards/alter-egos-iii/hermits/poepoeskizz-rare.ts
```diff
@@ -65,7 +65,7 @@ class PoePoeSkizzRare extends Card {
 					})
 					jumpscareAttack.addDamage(component.entity, 20)
 					jumpscareAttack.shouldIgnoreCards.push(query.card.entity(component.entity))
-					executeExtraAttacks(game, [jumpscareAttack])
+					attack.addNewAttack(jumpscareAttack)
```
Screaper91, line 68: The original form is correct because `attack` already finished executing after the pick request was fulfilled

common/cards/alter-egos-iii/hermits/poepoeskizz-rare.ts
```diff
@@ -37,7 +36,7 @@ class PoePoeSkizzRare extends Card {
 	override onAttach(game: GameModel, component: CardComponent, observer: ObserverComponent) {
 		const {player} = component
 
-		observer.subscribe(player.hooks.onAttack, (attack) => {
+		observer.subscribe(player.hooks.beforeAttack, (attack) => {
```
Screaper91, line 39: This should probably be in `afterAttack` to check if the attacker is still alive before picking to move

common/cards/season-x/hermits/skizzleman-rare.ts
```diff
@@ -33,58 +34,54 @@ class SkizzlemanRare extends Card {
...
-				)
-			) {
-				return
-			}
-
-			attackedEntities.add(attack.targetEntity)
+			if (attack.targetEntity) attackedEntities.add(attack.targetEntity)
 		})
 
 		observer.subscribe(player.hooks.onTurnEnd, () => {
```
Screaper91, line 74: Does not filter out the active Hermit and does not work when "Gas Light" is mocked

common/cards/season-x/hermits/skizzleman-rare.ts
```diff
@@ -33,58 +34,54 @@ class SkizzlemanRare extends Card {
...
 			}
+		})
 
+		observer.subscribe(opponentPlayer.hooks.onDefence, (attack) => {
 			// Status effect attacks have a special case because they happen at the end of the turn
-			if (attack.type === 'status-effect') {
-				if (attack.targetEntity && attackedEntities.has(attack.targetEntity)) return
-				attack.addDamage(component.entity, 20)
+			if (attack.type === 'status-effect' && attack.targetEntity) {
+				if (attackedEntities.has(attack.targetEntity)) return
```
Screaper91, line 66: Does not filter out active Hermit, may need clarification on interactions with Fire/Poison and Knockback/Egg
```suggestion
				if (attackedEntities.has(attack.targetEntity) || opponentPlayer.activeRowEntity === attack.targetEntity) return
```

Screaper91 (changes requested):
Both implementations still need work and possible clarification for edge cases

common/cards/alter-egos-iii/hermits/poepoeskizz-rare.ts
Lunarmagpie: Thats true. I think this should adjusted to do that.

common/cards/season-x/hermits/skizzleman-rare.ts
Lunarmagpie: I see. I believe I misunderstood the card.

common/cards/season-x/hermits/skizzleman-rare.ts
Lunarmagpie: I fixed this by changing mocked cards to delete themselves at the end of the turn instead of after attack.

Screaper91: This logic should probably be moved to a status effect since `onTurnEnd` will be removed if Skizz is knocked out by Thorns/Wolf

Lunarmagpie: I agree with that. I will convert this over to a status effect.

PR #755, opened 2024-07-29
Display Card Rarity

Lunarmagpie:
I have no idea why it didn't show this before?? Not showing rarity is far too confusing for new players so this has to be changed. 
I also removed the token costs because it was already displayed and didn't help with anything.


Lunarmagpie:
LGTM

PR #760, opened 2024-07-30
Change Strengths to readonly and reorganize Expansions

Benji42:
With these changes, the Strengths dict and the Expansions dict are both read only. This is important so people don't accidentally think they are supposed to modify these.
In addition, the expansions are now set up in a more intuitive way, and there also shouldn't be typing issues with them anymore, like there were in the past.


common/const/expansions.ts
```diff
@@ -0,0 +1,27 @@
+type ExpansionInfo = {
+	name: string
+}
```
Lunarmagpie, line 3: I think its overkill to have properties for the expansion when there is no other properties

common/const/strengths.ts
```diff
@@ -1,8 +1,22 @@
...
+	readonly builder: Array<TypeT>
+	readonly explorer: Array<TypeT>
+	readonly farm: Array<TypeT>
+	readonly miner: Array<TypeT>
+	readonly prankster: Array<TypeT>
+	readonly pvp: Array<TypeT>
+	readonly redstone: Array<TypeT>
+	readonly speedrunner: Array<TypeT>
+	readonly terraform: Array<TypeT>
+}
```
Lunarmagpie, line 15: This is the evil option, but there is nothing I like more than a good old metatype 
```suggestion
type Strengths = {
	readonly [Type in TypeT]: Array<TypeT>
}
```

Rvtar:
Now I actually have to learn how you set up the system.

Lunarmagpie:
It would be best if you use this type whenever possible https://www.typescriptlang.org/docs/handbook/utility-types.html#readonlytype

Rvtar:
I meant how I did my Weakness PR last time was messing with this since it's simpler to learn than how it currently works.

common/const/strengths.ts
```diff
@@ -1,8 +1,12 @@
 import {TypeT} from '../../common/types/cards'
 
-export const STRENGTHS: Record<TypeT, Array<TypeT>> = {
-	balanced: [],
+type strengths = {
```
Lunarmagpie, line 3: This is a type so it must be capital

Lunarmagpie (changes requested):

common/const/strengths.ts
Benji42: Fixed

Lunarmagpie (approved):

PR #764, opened 2024-07-30
Fix errors in logs

Screaper91:
- Shield now logs the name of Hermit and row # when broken
- Multi-target attacks show correct target damaged by each sub-attack
- Rare Goatfather and Spooky Stress no longer attack empty rows
- Anvil no longer ignores rows with Armor Stand below the active


Lunarmagpie:
I tested this and it works great. The battle log is @Benji42's specialty so I would like him to review this before we merge this.

Benji42 (approved):

PR #766, opened 2024-07-31
Princess gem fix

Benji42:
Change Princess Gem's description/behavior to be able to properly block status effects


Lunarmagpie (approved):
LGTM

PR #770, opened 2024-07-31
Skip beef coinflip if there is no status effects

Lunarmagpie:


Lunarmagpie:
I am leaving the descriptions unchanged for now. These can be updated in the future.

PR #771, opened 2024-07-31
Update api to work with ecs

Tyrannicodin:


common/components/query/player.ts
```diff
@@ -17,3 +17,9 @@ export function hasStatusEffect(
 		return !!player.hasStatusEffect(statusEffect)
 	}
 }
+
+export function id(id: string): ComponentQuery<PlayerComponent> {
+	return (_game, player) => {
+		return player.id === id
+	}
+}
```
Lunarmagpie, line 25: I would prefer if you did not write a combinator for this.
Normally it would be okay, but I am trying to decouple player models from their IDs. Merging this than undoing it would end up being a bit of a waste of time so I would prefer if you used my alternate suggestion for the api.

server/src/api.ts
```diff
@@ -3,37 +3,46 @@ import {GameModel} from 'common/models/game-model'
...
 								return {
 									createdTime: g.createdTime,
 									id: g.id,
 									code: g.code,
-									playerIds: g.getPlayerIds(),
-									playerNames: g.getPlayers().map((p) => p.name),
+									players: g.getPlayers().map((p) => {
+										const playerState: PlayerStateT = p.getPlayerInfo() as PlayerStateT
+										playerState.lives = g.components.filter(PlayerComponent, query.player.id(p.id))[0].lives
+										return playerState
```
Lunarmagpie, line 44: Are you able to do this?
```suggestion
									players: g.combinators.filter(PlayerComponent).map((player) => {
									     // The properties you need to send to the bot
									     lives: player.lives,
									}
```
It is not a good idea to monkeypatch objects. Please make a new dictionary instead.

Lunarmagpie (changes requested):
I suggest using longer variable names in the code. Single letter names are confusing for people who are not yet familiar with the codebase.

server/src/api.ts
Tyrannicodin: I was unaware I could do this, and agree this is much nicer

PR #773, opened 2024-07-31
Fix Jamm Bug with Logging Row Index

Lunarmagpie:
Its funny how saying "Jamm Bug" is the fastest way to explain what this PR does


Tyrannicodin (approved):
I thought ecs was supposed to remove that forever lol

PR #774, opened 2024-07-31
Move deck select screen to own file

Screaper91:
Fixes issue where selecting a new deck resets scroll position of deck list


Lunarmagpie (approved):
LGTM

PR #775, opened 2024-08-01
Add sound effects

Lunarmagpie:
Adds sound effects to the game:
- Place in item slot sound for playing cards
- Remove from item slot sound for when cards are removed from the board
- Item frame rotate sound for changing active hermit
- Damage sound
- Piston Sound
- Chorus Fruit Sound

The damage sound and piston sound feel quite loud. This needs adjustments from somebody who understands audio.


Lunarmagpie:
This isn't the right way to do this. I have to start over.

PR #789, opened 2024-08-03
Completely disconnect game from player models

Lunarmagpie:
This is a fun one.
After this, we should finally be able to create games without player models.

I added a test to show how this is possible. I did not investigate redux saga testing yet.


Benji42:
Is this ready for review?

Lunarmagpie:
Should be ready for review now

PR #793, opened 2024-08-04
Add Spectators

alicetouhou:
This pull requests adds a basic implementation of spectators. There is one limitation, there is no way to control if spectators are allowed to chat. Players also cannot see their spectator code during the match. Spectators can leave and join games or the queue at any time with no issue.


Lunarmagpie:
I think this PR is ready.

Lunarmagpie:
Before I merge this I am going to investigate writing some tests for the matchmaking saga. The code is pretty confident and I don't trust myself to have caught every edge case.

alicetouhou:
Just add more tests 🔥🔥🔥

PR #795, opened 2024-08-04
Fix XB Not Dealing Type Advanage Damage to Targets With Armor

Lunarmagpie:


Benji42:
This seems like a good fix, but issues like this make me think we should remove weakness attacks as a concept entirely

Benji42 (approved):

PR #797, opened 2024-08-04
Fix Wrong Player Being Marked as Dead

Lunarmagpie:
Fixes a bug where the wrong player is being shown as dead. This should also fix firebase.


Benji42:
FINALLY NO MORE TS IGNORE
LETSGOOOO

Lunarmagpie:
> FINALLY NO MORE TS IGNORE LETSGOOOO

The `any` is not much better lol.

PR #799, opened 2024-08-04
Change sound settings words to be more obvious

Benji42:
Before, I think it was a bit confusing how muting worked. I've changed it to be more obvious about which option is muted and which is unmuted


Lunarmagpie:
LGTM

PR #803, opened 2024-08-05
Improve curse of binding status effect

Benji42:
- Added an icon to make the effect more easily visible
- Fixed issue where you could still switch with Chorus Fruit, Cub, or Tango


Lunarmagpie:
Is this not a system status effect?

Benji42:
oh yeah it is, fixed

PR #807, opened 2024-08-05
Add Types To Redux

Lunarmagpie:
This PR adds types to redux. I added types for the client local messages and communication between the client and the server. I additionally refactored the server to use selectors because that makes the control flow easier to follow.

This pull request is NOT intended to be a final version of all of this. This part of the code will need to be reorganized and refactored. The goal of this PR is to add types to make that doable.


Lunarmagpie:
I believe this is done and I am happy with the code. I just need more testing at the moment to verify everything works.

Benji42 (approved):

PR #808, opened 2024-08-06
Fix Potatoboy's ability triggering with secondary instead of primary

Rvtar:


common/cards/alter-egos/hermits/potatoboy-rare.ts
```diff
@@ -44,7 +44,7 @@ class PotatoBoyRare extends Card {
 		const {player} = component
 
 		observer.subscribe(player.hooks.onAttack, (attack) => {
-			if (!attack.isAttacker(component.entity) || attack.type !== 'secondary')
+			if (!attack.isAttacker(component.entity) || attack.type == 'secondary')
```
Lunarmagpie: Its best to use `===`. @Benji42 Would kill me if I didn't make you change this lol.

Lunarmagpie (approved):

common/cards/alter-egos/hermits/potatoboy-rare.ts
Rvtar: I changed the change to be primary instead when I realized it might trigger on effect attacks.

Lunarmagpie:
Thank you!

PR #809, opened 2024-08-06
Fix anvil and hyperspace logs

Screaper91:
- Anvil shows correct damage preview when there is not a hermit opposite your active
- Fixes Hyperspace log when it does not damage a hermit
- Fixes Anvil logs when it does no damage or there is not a hermit opposite your active

### DIFF
--- a/common/cards/alter-egos/hermits/renbob-rare.ts
+++ b/common/cards/alter-egos/hermits/renbob-rare.ts
@@ -9,6 +9,8 @@ import Card from '../../base/card'
 import {hermit} from '../../base/defaults'
 import {Hermit} from '../../base/types'
 
+const EMPTY_ROW = '$o$bINVALID_VALUE$$'
+
 class RenbobRare extends Card {
 	props: Hermit = {
 		...hermit,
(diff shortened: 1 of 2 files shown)


common/cards/alter-egos/hermits/renbob-rare.ts
```diff
@@ -56,6 +58,14 @@ class RenbobRare extends Card {
...
 				)?.entity || null,
 			)
+			attack.updateLog((values) => {
+				if (values.previousLog === undefined) return ''
+				return values.target === EMPTY_ROW
+					? values.previousLog
+							.replace(`${values.target} `, '')
+							.replace(`for ${values.damage} damage`, 'and missed')
+					: values.previousLog
+			})
```
Lunarmagpie, line 68: I don't think this is the right approach. It would be better to check if the attack doesn't have a target  and write a log for that case instead of cleaning up for the other broken stuff.

Benji42:
For hyperspace, I think a better solution is to pass the attack or game into the log function. With that, we can have an if statement in the function to give a different log if the attack has no target. Pretty sure this would be simpler. We can do this for the default Hermit attack log, even, and have something like "HERMIT attacked with ATTACK and missed" if the attack has no target or deals zero damage from multiplication.

Screaper91:
This was functional, but the approach for Hyperspace was not as clean as needed to be maintainable going forward

PR #810, opened 2024-08-06
Fix Timeskip effect query

Screaper91:
Timeskip can no longer flip a coin to skip a turn if Clock was played on the same turn or the player's previous turn


Lunarmagpie (approved):

Lunarmagpie:
LGTM

PR #812, opened 2024-08-08
Settings types

Benji42:
Types settings so they are much easier to work with and to prevent errors
Also fixes a bug where toast sounds would not play due to referencing a setting that did not exist


client/src/logic/local-settings/local-settings-reducer.ts
```diff
@@ -1,38 +1,58 @@
...
+	confirmationDialogs: true,
+	showChat: false,
+	showBattleLogs: false,
+	showAdvancedTooltips: true,
+	chatPosition: {x: 0, y: 0},
+	chatSize: {w: 0, h: 0},
+	panoramaEnabled: true,
+	panorama: 'hermit-hill',
+	gameSide: 'Left',
+	minecraftName: 'alex',
```
Lunarmagpie, line 36: i would change some of these to be like `profanityFilterEnabled`

Benji42: Planning on doing this in a separate update, feel it's out of scope

PR #817, opened 2024-08-08
Redo Definitions Creation, again...

Lunarmagpie:
I believe in my last attempt to fix the cards I made a fatal error. It was far too complicated. We can remove the whole `CardProps` thingy and completely remove classes from the equation. The old system relied on many typescript features to work, while with this solution we only need structural typing. This solution is much easier to understand and should be very simple for future maintainers to work around and makes it clear how to extend the system in an easy way.


Benji42:
To me, this looks a lot better, but I would also like someone else's feedback on this change before we merge

PR #819, opened 2024-08-08
Rework settings to be more intuitive for develops and players

alicetouhou:
- Renames some settings, and changes all of them that used `disabled` to `enabled`
- Fixes a bug where confirmation dialogues would follow the opposite of what the setting was set to
- Changes "Ingame" to "In-Game" on the settings screen, since the spelling was wrong before
- Makes other changes for visual consistency
- Adds a button to reset settings

Unfortunately, this will bring some settings back to the default again but it is possible for us to add a button in "Data Management" to fix this if needed

### DIFF
--- a/client/src/app/app.tsx
+++ b/client/src/app/app.tsx
@@ -65,10 +65,10 @@ function App() {
 		return (
 			<Background
 				panorama={settings.panorama}
-				disabled={!settings.panoramaEnabled}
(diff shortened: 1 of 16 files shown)


Lunarmagpie:
its better to end with `enabled` instead of starting with `enable`.

Lunarmagpie:
LGTM

PR #820, opened 2024-08-08
Fix stats menu and add rounding to winrate

alicetouhou:


client/src/app/main-menu/settings.tsx
```diff
@@ -16,7 +16,11 @@ function Settings({setMenuSection}: Props) {
 	const dispatch = useMessageDispatch()
 	const stats = useSelector(getStats)
 	const settings = useSelector(getSettings)
-	const totalGames = Object.values(stats).reduce((a, b) => a + b, 0)
+
+	const totalGames = Object.values(stats).reduce((a, b) => {
+		if (typeof b === 'number') return a + b
+		return a
+	}, 0)
```
Lunarmagpie, line 24: This is not good. just do `totalGames = stats.w + stats.l + stats.fw + stats.fl + stats.t`

alicetouhou: true, would be easier to understand

Lunarmagpie (approved):

PR #821, opened 2024-08-08
Fix turtle shell

alicetouhou:


common/cards/alter-egos/effects/turtle-shell.ts
```diff
@@ -23,29 +23,37 @@ class TurtleShell extends Card {
...
+		observer.subscribe(
+			player.hooks.onActiveRowChange,
+			(_oldActiveHermit, newActiveHermit) => {
+				const hermitCard = game.components.find(
+					CardComponent,
+					query.card.isHermit,
+					query.card.row(query.row.hasCard(component.entity)),
+				)
+
+				if (hermitCard && newActiveHermit === hermitCard) {
```
Lunarmagpie, line 42: You should compare the card entities and not the objects. Unfortunately this code is wrong and technically works but i can't force it to error.

Lunarmagpie (changes requested):

common/cards/alter-egos/effects/turtle-shell.ts
alicetouhou: Oops, fixed

Lunarmagpie (approved):

PR #834, opened 2024-08-11
Fix gaslight activating when it shouldn't have

alicetouhou:
Gaslight still activates when an attack does 0 damage. If an attack does zero damage, after armor, gaslight should not activate according to the description


Lunarmagpie:
It looks like you are checking the damage to me. I don't think its a good idea to do that.

Lunarmagpie (approved):

Lunarmagpie (approved):

PR #835, opened 2024-08-11
Call onActiveRowChange with no previous active row

Screaper91:
Fixes issues where hook was not called when players selected a new active row after their active hermit was knocked out


Lunarmagpie (approved):
LGTM

PR #836, opened 2024-08-11
Introduce priority hooks to fix ordering issues

Screaper91:
- Lightning Rod can once again be ignored by attacks
- rare Dwarf Impulse and xB show what attach effects are ignored with a status effect
- Fixes single use requests occurring before hermit attack requests
- Fixes incorrect single use logs when stolen by Trap Hole
- Fixes rare PoultryMan retrieving Eggs stolen by Trap Hole
- Fixes Egg not dealing damage to selected hermits
- Fixes Sheep Stare triggering on Thorns, Wolf, and Ender Pearl
- Fixes being able to skip attacking an AFK hermit for Betrayed by only attacking with a single use card
- Fixes rare Doc dealing 120 damage when World Eater and Invisibility flip tails
- Fixes rare Hotguy doubling all Bow damage if made AFK after using secondary
- Fixes rare interactions with Thorns, Wolf, Shield, Totem, and Deathloop


common/cards/alter-egos-iii/hermits/kingjoel-rare.ts
```diff
@@ -45,52 +45,55 @@ class KingJoelRare extends Card {
...
 				return
-			if (!game.components.exists(SlotComponent, firstPickCondition)) return
-			if (!game.components.exists(SlotComponent, secondPickCondition)) return
+			const firstPickCondition = query.every(
+				query.slot.opponent,
+				query.not(query.slot.rowIs(player.opponentPlayer.activeRowEntity)), // Specified in case Knockback or Egg is used
+				query.slot.item,
+				query.not(query.slot.empty),
+				query.slot.row((_game, row) => !!row.health),
+			)
```
Lunarmagpie, line 67: I don't think the comment is necessary here
```suggestion
			const firstPickCondition = query.every(
				query.slot.opponent,
				query.not(query.slot.active),
				query.slot.item,
				query.not(query.slot.empty),
				query.slot.row((_game, row) => !!row.health),
			)
```

Lunarmagpie:
I don't think its necessary to move the `firstPickCondition` variable into the onAttack hook scope. I think it would be better to have it defined before `secondPickCondition` like it was before.

Lunarmagpie:
I think it would be simpler to keep King Joel with the `onAttack` hook. This matches his description better. Additionally, then `egg`, `chorus`, etc should be able to use `afterAttack` to make sure they trigger after Joel.

Lunarmagpie:
I think the best way to solve this issue will be implementing the hook priority system we discussed in developer general. If you want to implement that in this PR I think that would be a good idea.

Lunarmagpie:
Now that we have unit tests I think that would fit into this PR quite well. Once we have time to discuss this again I'd like to make sure that is implemented.

common/components/observer-component.ts
```diff
@@ -50,6 +51,24 @@ export class ObserverComponent {
 		this.hooks.push(hook)
 	}
 
+	/** Subscribe to a priority hook with this observer. This hook will be removed when the observer is destroyed.
+	 * Cards and status effects will destroy their own observers, so you as the user do not need to
+	 * worry about this!
+	 * If you are looking for a hook that will be called after the observer is destroyed (the card is removed
+	 * from the board), please use a status effect instead.
+	 */
+	public subscribeWith<
```
Lunarmagpie, line 60: I think the function name `subscribeWithPriority` would be a little clearer

Lunarmagpie:
Thanks for the work! I will look this over as soon as I can.

common/types/priorities.ts
```diff
@@ -0,0 +1,94 @@
...
+	{stage: 'PUSH_IGNORE_CARDS'},
+	{stage: 'HERMIT_SET_TARGET'},
+	{stage: 'HERMIT_CHANGE_TARGET'},
+	{stage: 'LIGHTNING_ROD_REDIRECT'},
+	{stage: 'TARGET_BLOCK_REDIRECT'},
+	{stage: 'HERMIT_MODIFY_DAMAGE'},
+	{stage: 'EFFECT_MODIFY_DAMAGE'},
+	{stage: 'HERMIT_APPLY_ATTACK'},
+	{stage: 'APPLY_SINGLE_USE_ATTACK'},
+] as const)
```
Lunarmagpie, line 49: This style of dictionary doesn't work with the LSP goto definition and goto references features.
It will be better to do this:

```
const priorities = createPriorityDictionary({
	'HELLO_WORLD': null,
	'SOMETHING_ELSE': null,
})

```

I did this for client and server references if you want to look at an example.

Lunarmagpie:
If you can adjust the comments to be more descriptive and look like they do not list everything I would appreciate that. I imagine these comments will be forgotten about when new cards are added so I don't want anyone to be confused in a year's time.

Lunarmagpie:
This is a bit of a petty request but if you could change the names of the steps to have the verb at the end I think its a little nicer. `EFFECT_REMOVE_DAMAGE -> EFFECT_DAMAGE_REMOVE` for example. It makes it easier to see related items because the text lines up exactly until the last verb.

common/types/priorities.ts
```diff
@@ -0,0 +1,94 @@
...
+ * @member HERMIT_CHANGE_TARGET: Betrayed effect, Sheep Stare effect
+ * @member LIGHTNING_ROD_REDIRECT
+ * @member TARGET_BLOCK_REDIRECT
+ * @member HERMIT_MODIFY_DAMAGE: Hermits changing the damage of an attack or adding sub-attacks
+ * @member EFFECT_MODIFY_DAMAGE: Invisibility effects
+ * @member HERMIT_APPLY_ATTACK: Hermits applying their effect to the board
+ * @member APPLY_SINGLE_USE_ATTACK: Single use attack cards: Bow, Swords, etc.
+ */
+export const beforeAttack = createPriorityDictionary([
+	{stage: 'PUSH_IGNORE_CARDS'},
```
Lunarmagpie, line 40:
```suggestion
	{stage: 'IGNORE_CARDS'},
```
The push in this really confused me for a minute so I think its clearer without it.

tests/game/helsknight-rare.test.ts
```diff
@@ -0,0 +1,49 @@
...
+						game.chat.find(
+							(message) =>
+								message.message.TYPE === 'ListNode' &&
+								message.message.nodes.find(
+									(node) =>
+										node.TYPE === 'PlaintextNode' &&
+										node.text.includes('backlash damage'),
+								),
+						)?.sender.id,
+					).toBe(game.currentPlayerEntity)
```
Lunarmagpie, line 39: I don't think this is a good way to check if TNT was used. I think having a dependency on the logging system in the game tests will make this very confusing to fix if we ever refactor logs or the TNT log changes. I would prefer if you checked the HP of the hermits.

tests/game/kingjoel-rare.test.ts
```diff
@@ -0,0 +1,126 @@
...
+						query.slot.rowIndex(1),
+						query.slot.index(0),
+					)
+
+					expect(
+						game.components.find(
+							CardComponent,
+							query.card.is(BalancedItem),
+							query.card.currentPlayer,
+						),
```
Lunarmagpie, line 102: I think the change expresses the intent of the test better.
```suggestion
						game.components.find(
							CardComponent,
							query.card.is(BalancedItem),
							query.card.currentPlayer,
							query.card.attached,
						),
```

tests/game/kingjoel-rare.test.ts
```diff
@@ -0,0 +1,126 @@
...
+					).toBeTruthy()
+				},
+			},
+			{startWithAllCards: true, noItemRequirements: true, forceCoinFlip: true},
+		)
+	})
+
+	test('Test Steal with Egg', () => {
```
Lunarmagpie, line 61:
```suggestion
	test('Test Steal pick request runs before Egg pick request.', () => {
```

Lunarmagpie (changes requested):
Can you change the Totem Of Undying tests that use scar but do not test deathloop to a different Hermit?

tests/game/helsknight-rare.test.ts
Screaper91: This check was to verify the current player sent the attack log instead of the opponent player (the new owner of TNT), I can remove this test and system dependency now that the sender is fixed

Lunarmagpie: I see. I'm not sure I see this as a valuable test so I prefer if you removed it.

common/components/observer-component.ts
```diff
@@ -50,6 +51,24 @@ export class ObserverComponent {
 		this.hooks.push(hook)
 	}
 
+	/** Subscribe to a priority hook with this observer. This hook will be removed when the observer is destroyed.
+	 * Cards and status effects will destroy their own observers, so you as the user do not need to
+	 * worry about this!
+	 * If you are looking for a hook that will be called after the observer is destroyed (the card is removed
+	 * from the board), please use a status effect instead.
+	 */
```
Lunarmagpie, line 59: Copying the entire comment will almost certainly lead to documentation rot.
```suggestion
	/** Subscribe to a priority hook with this observer. Functions similarly to `subscribe`. */
```

Lunarmagpie:
The core stuff in this PR looks awesome! I really like the design choice of not making `PriorityHook` inherit from the regular hook class. Its gonna be so much easier to change anything relating to it if we ever need to. I am ready to merge this right now because your implementation is very easy to change if needs be so theres no use overthinking it.

Lunarmagpie (approved):

tests/game/helsknight-rare.test.ts
Lunarmagpie: I actually just meant the one part with the TNT log check - I added the rest of the test back because I think its really useful.

PR #840, opened 2024-08-12
Add Safe Call Function To Prevent Errors Bubling Up To Root

Lunarmagpie:


Lunarmagpie:
Although I don't really like this, it needs to be merged urgently. We can easily change this later.

PR #843, opened 2024-08-12
Improve Framework For Card Unit Tests

Lunarmagpie:
This PR adds the ability to test cards and some basic functions to make it easier. I think this is a good start. We can expand this as we need to test more things.
I added a test for Clock, Loyalty, and Impulse. There should be enough examples in there for anyone to make the test they want.


Lunarmagpie:
I have tested that general behavior with cards, status effects, and turn actions works as intended. I believe this should be good to merge.

PR #862, opened 2024-08-23
Fix Golden Axe with Geminislay

Screaper91:
Golden Axe now applies a status effect that ignores the opponent's attach slot for the entire turn.
- Fixes Gem's second single use attack triggering opponent's attach effect if Golden Axe was used first


Lunarmagpie (approved):
Thanks for the work and very well done! I would like to improve our test coverage now that tests are possible, so if you can also write that also checks that the status effect expires at the right time? Or add it to the Gem one, your call on what you think is the best.

Lunarmagpie (approved):

PR #864, opened 2024-08-25
Fix "Peace Out" timeout query

Screaper91:
- Fixes Peace Out being able to change opponent's active row to invalid values if request is timed out

### DIFF
--- a/common/components/player-component.ts
+++ b/common/components/player-component.ts
@@ -258,6 +258,12 @@ export class PlayerComponent {
 		// Can't change to existing active row
 		if (newRow === currentActiveRow) return false
 
+		// Can't change to opponent's row
+		if (newRow.playerId !== this.entity)
+			throw new Error(
+				"Should not be able to change to another player's row to make active",
+			)
+
 		// Call before active row change hooks - if any of the results are false do not change
 		if (currentActiveRow) {
 			let oldHermit = currentActiveRow.getHermit()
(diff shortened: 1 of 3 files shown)


common/components/player-component.ts
```diff
@@ -258,6 +258,12 @@ export class PlayerComponent {
 		// Can't change to existing active row
 		if (newRow === currentActiveRow) return false
 
+		// Can't change to opponent's row
+		if (newRow.playerId !== this.entity)
+			throw new Error(
+				"Should not be able to change to another player's row to make active",
+			)
```
Lunarmagpie, line 265: I would prefer:
```suggestion
		assert(newRow.playerId === this.entity, "Should not be able to change to another player's row to make active")
```

Lunarmagpie:
Looks good besides for my one comment

common/components/player-component.ts
```diff
@@ -258,6 +258,12 @@ export class PlayerComponent {
 		// Can't change to existing active row
 		if (newRow === currentActiveRow) return false
 
+		// Can't change to opponent's row
```
Lunarmagpie, line 261: It would be better not to have a comment before an assert. The assertion here already tells us what the comment does and its MORE POWERFUL :rocket:  because if anything changes in the game and our comment is wrong, the assertion will then fail.
Regular comments don't have this power so its best to use them only if you can't assert the result or printing out what the comment says in a log doesn't make sense either.

Lunarmagpie (approved):

PR #870, opened 2024-09-02
Fix issue with Renbob's log showing undefined when there's no valid target

alicetouhou:
This also gives a different description now when the attack misses from a multiplier, I think to the player this will be more intuitive.

### DIFF
--- a/common/cards/default/hermits/rendog-rare.ts
+++ b/common/cards/default/hermits/rendog-rare.ts
@@ -61,7 +61,7 @@ const RendogRare: Hermit = {
 		newAttack.updateLog((values) => {
 			if (
 				values.attack.getDamageMultiplier() === 0 ||
-				values.attack.target?.getHermit() === null
+				!values.attack.target?.getHermit()
 			) {
 				return `${values.attacker} ${values.coinFlip ? values.coinFlip + ', then ' : ''} attacked with ${values.attackName} and missed`
 			}
--- a/common/cards/default/hermits/rendog-rare.ts
+++ b/common/cards/default/hermits/rendog-rare.ts
@@ -61,9 +61,9 @@ const RendogRare: Hermit = {
 		newAttack.updateLog((values) => {
 			if (
 				values.attack.getDamageMultiplier() === 0 ||
-				values.attack.target === null
+				values.attack.target?.getHermit() === null
 			) {
-				return `${values.attacker} ${values.coinFlip ? values.coinFlip + ', then ' : ''} attacked and missed`
+				return `${values.attacker} ${values.coinFlip ? values.coinFlip + ', then ' : ''} attacked with ${values.attackName} and missed`
 			}
 			return `${values.attacker} ${values.coinFlip ? values.coinFlip + ', then ' : ''} attacked ${
 				values.target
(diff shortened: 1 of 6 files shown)


Lunarmagpie:
Poultryman test failed

server/src/routines/game.ts
```diff
@@ -714,10 +714,9 @@ function* checkDeckedOut(game: GameModel) {
 		game.settings.unlimitedCards
 	)
 		return []
-	return [game.currentPlayer, game.opponentPlayer].flatMap((player) => {
-		if (player.deckedOut) return [player]
-		return []
-	})
+	return [game.currentPlayer, game.opponentPlayer].filter(
+		(player) => player.deckedOut,
+	)
```
Lunarmagpie, line 719: I would prefer if you didn't touch this here since screaper is working on similar things ATM it might cause a conflict

common/cards/default/hermits/rendog-rare.ts
```diff
@@ -58,12 +58,17 @@ const RendogRare: Hermit = {
...
-		newAttack.updateLog(
-			(values) =>
-				`${values.attacker} ${values.coinFlip ? values.coinFlip + ', then ' : ''} attacked ${
-					values.target
-				} with $v${mockedAttack.hermitName}'s ${attackName}$ for ${values.damage} damage`,
-		)
+		newAttack.updateLog((values) => {
+			if (
+				values.attack.getDamageMultiplier() === 0 ||
+				values.attack.target?.getHermit() === null
```
Lunarmagpie, line 64: What is `values.attack.target?.getHermit()` is `undefined`

common/models/battle-log-model.ts
```diff
@@ -242,6 +242,7 @@ export class BattleLogModel {
 				damage: `$b${subAttack.calculateDamage() + weaknessDamage}hp$`,
 				defaultLog: this.generateEffectEntryHeader(singleUse),
 				coinFlip: this.generateCoinFlipMessage(attack, coinFlips),
+				attack: subAttack,
```
Lunarmagpie, line 245: Where are you reading this property?

server/src/routines/game.ts
alicetouhou: Kinda confused why this changed because I didn't touch it.

common/cards/default/hermits/rendog-rare.ts
alicetouhou: According to the types this shouldn't be possible but I can change it to a !

common/models/battle-log-model.ts
alicetouhou: the default hermit attack log, and Ren and Cleo

common/cards/default/hermits/rendog-rare.ts
Lunarmagpie: Don't change it to a !. Write an assertion if you are confident it can't happen.

alicetouhou: nah i was wrong it can be undefined lol

Lunarmagpie (approved):

Lunarmagpie (approved):

PR #872, opened 2024-09-03
Improve calculation of game winner

Screaper91:
Decking out is now only considered if both players are still alive
- Fixes running out of cards causing losses to be ties or vice versa
- Improves `getGamePlayerOutcome` to support future PvE and spectator games

### DIFF
--- a/server/src/routines/game.ts
+++ b/server/src/routines/game.ts
@@ -636,8 +636,11 @@ export function* turnSaga(game: GameModel) {
 	if (game.state.turn.turnNumber > 2) {
 		const turnStartDeadPlayers = yield* call(checkHermitHealth, game)
 		if (turnStartDeadPlayers.length) {
-			game.endInfo.reason =
-				turnStartDeadPlayers[0].lives <= 0 ? 'lives' : 'hermits'
+			game.endInfo.reason = turnStartDeadPlayers.every(
+				(deadPlayer) => deadPlayer.lives <= 0,
+			)
+				? 'lives'
+				: 'hermits'
 			game.endInfo.deadPlayerEntities = turnStartDeadPlayers.map(
 				(player) => player.entity,
 			)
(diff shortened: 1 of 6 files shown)


server/src/utils/win-conditions.ts
```diff
@@ -18,40 +42,37 @@ export const getGameOutcome = (game: GameModel, endResult: any) => {
...
-export const getWinner = (game: GameModel, endResult: any) => {
+export const getWinner = (game: GameModel, endResult: EndResult) => {
 	if (Object.hasOwn(endResult, 'timeout')) return null
 	if (Object.hasOwn(endResult, 'forfeit')) {
-		return getOpponentId(game, endResult.forfeit.playerId)
+		return getOpponentId(game, endResult.forfeit!.playerId)
 	}
 	if (Object.hasOwn(endResult, 'playerRemoved')) {
-		return getOpponentId(game, endResult.playerRemoved.player.id)
+		return getOpponentId(game, endResult.playerRemoved!.player.id)
```
Lunarmagpie, line 75: I would prefer if you used an assert statement instead of the bang operator. This section of the code is pretty complex so when something does go wrong I want good feedback there.

server/src/routines/game.ts
```diff
@@ -668,14 +671,9 @@ export function* turnSaga(game: GameModel) {
...
-	let deadPlayers: Array<PlayerComponent> = []
-	deadPlayers.push(...(yield* call(checkHermitHealth, game)))
-	deadPlayers.push(...(yield* call(checkDeckedOut, game)))
-
+	const deadPlayers: PlayerComponent[] = yield* call(checkHermitHealth, game)
 	if (deadPlayers.length) {
-		if (deadPlayers[0].deckedOut) {
-			game.endInfo.reason = 'cards'
-		} else if (deadPlayers[0].lives <= 0) {
+		if (deadPlayers.every((player) => player.lives <= 0)) {
```
Lunarmagpie, line 676: Shouldn't this be `deadPlayers.some()`? Also the same thing for the other occurrence.

Lunarmagpie (changes requested):

server/src/routines/game.ts
Screaper91: If the game ends in a tie where one player is out of lives and the other is out of hermits, should they both see the message "You ran out of `lives`" or "You ran of `hermits`"? (current patch shows whatever `playerOne` ran out of)

Lunarmagpie: Ok I see! It doesn't really matter to me so I like this implementation.

Lunarmagpie:
Testing and this works!

PR #874, opened 2024-09-05
Cleo Puppetry fixes

Screaper91:
- Fixes Amnesia disabling Cleo's primary the turn after using Puppetry to mock another hermit's primary
- Fixes exploit allowing Puppetry to use the effects of multiple attacks at once
- Fixes Boomer Bdubs adding duplicate pick requests when secondary is used with Bow/Crossbow

### DIFF
--- a/server/src/routines/turn-actions.ts
+++ b/server/src/routines/turn-actions.ts
@@ -29,6 +29,7 @@ function getAttack(
 	if (!creator.isHermit()) return []
 
 	const nextAttack = creator.props.getAttack(game, creator, hermitAttackType)
+	creator.player.updateLastUsedHermitAttack(hermitAttackType)
 
 	if (nextAttack) attacks.push(nextAttack)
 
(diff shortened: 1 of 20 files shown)


Lunarmagpie:
Can you move the boomer bdubs changes to a different pull request? I am happy to merge those changes right now but I need a little more time to look at the Cleo/GA changes.

Lunarmagpie:
Im not sure i'm happy with the `lastHermitAttack` being moved to the player component class because it is only used by one card. What was the reason you needed to move this?

server/src/routines/turn-actions.ts
```diff
@@ -29,6 +29,7 @@ function getAttack(
 	if (!creator.isHermit()) return []
 
 	const nextAttack = creator.props.getAttack(game, creator, hermitAttackType)
+	creator.player.updateLastUsedHermitAttack(hermitAttackType)
```
Screaper91, line 32: This is the only place that the selected attack type can be forwarded when the attack action is executed. In order for Amnesia to block Puppetry/Role Play when they mock a primary, attack type must read here instead of `attack.type`. Alternative hooks to `beforeAttack` are not usable because attacks can be canceled after running `getAttackRequests` and `getAttack` does not take a `HermitAttackType` parameter

Lunarmagpie: That makes sense. I don't really think this is a refactor worthy thing to worry about so I'm happy with this.

Lunarmagpie (approved):

PR #875, opened 2024-09-05
Fix Poultry Man returning Egg late

Screaper91:
- Fixes rare Poultry Man secondary not returning Egg until `player.hooks.afterAttack` was run again

_This caused `poultryman-rare.test.ts` to fail when Egg flipped tails and did not attack target hermit, Egg was not in hand yet._


Lunarmagpie (approved):

Lunarmagpie:
I added a test for rolling tails to this PR so we have both cases in the future.

PR #876, opened 2024-09-05
Attack request fixes

Screaper91:
Includes changes from #874 not part of Puppetry behavior:
- rare Boomer Bdubs no longer adds duplicate requests after flipping heads for "Watch This"
- Betrayed effect no longer remembers selected AFK hermit after opponent cancels their attack


Lunarmagpie:
LGTM

PR #877, opened 2024-09-06
Allow Invisibility and Aussie Ping To Affect All Attacks Created By Hermits

Lunarmagpie:
Attempts to fix the consistency issues between Invisibility and Aussie Ping noted by Jamm.

> so because of this,  i've been playing around with Invis which I don't use and I noticed that Invis
> - can double/block lLamadad's additional damage (+40)
> - can double/block Goatfather's +30 but not the +10 to hermits below the active
> - can't double/block any of the Skizzs additional damage
> - can't double/block S. Stress +10 to AFK
>
> compared to Pearl
> - Pearl can block lLamadad's additional damage (+40)
> - Pearl can block Goatfather's +30 and +10 to hermits below the active
> - Pearl can't block any of Skizzs additional damage
> - Pearl can block S. Stress +10 to AFK

This PR makes Invisibility Potion and Pearl trigger in all of these scenarios.

Add tests for both regular skizz''s to ensure that the tests with pearl and invisibility are not relying on incorrect behavior.

### DIFF
--- a/common/status-effects/invisibility-potion.ts
+++ b/common/status-effects/invisibility-potion.ts
@@ -26,6 +26,7 @@ export const InvisibilityPotionHeadsEffect: StatusEffect<PlayerComponent> = {
 			beforeAttack.MODIFY_DAMAGE,
 			(attack) => {
 				if (!attack.isType('primary', 'secondary')) return
+				multipliedDamage = true
 				attack.multiplyDamage(effect.entity, 0)
 			},
 		)
@@ -58,6 +59,7 @@ export const InvisibilityPotionTailsEffect: StatusEffect<PlayerComponent> = {
 			beforeAttack.MODIFY_DAMAGE,
 			(attack) => {
 				if (!attack.isType('primary', 'secondary')) return
+				multipliedDamage = true
 				attack.multiplyDamage(effect.entity, 2)
 			},
 		)
--- a/common/status-effects/invisibility-potion.ts
+++ b/common/status-effects/invisibility-potion.ts
@@ -4,7 +4,7 @@ import {
 	StatusEffectComponent,
 } from '../components'
 import {GameModel} from '../models/game-model'
-import {beforeAttack} from '../types/priorities'
+import {beforeAttack, onTurnEnd} from '../types/priorities'
 import {StatusEffect, systemStatusEffect} from './status-effect'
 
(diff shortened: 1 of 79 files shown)


common/status-effects/invisibility-potion.ts
```diff
@@ -21,10 +21,16 @@ export const InvisibilityPotionHeadsEffect: StatusEffect<PlayerComponent> = {
...
 			(attack) => {
 				if (!attack.isType('primary', 'secondary')) return
 				attack.multiplyDamage(effect.entity, 0)
+			},
+		)
+		observer.subscribeWithPriority(
+			player.opponentPlayer.hooks.onTurnEnd,
+			onTurnEnd.ON_STATUS_EFFECT_TIMEOUT,
+			() => {
 				effect.remove()
```
Screaper91, line 34: Invisibility status effects should be persistent until they modify an attack
```suggestion
				if (multipliedDamage) effect.remove()
```

tests/game/invisibility-potion.test.ts
```diff
@@ -0,0 +1,195 @@
...
+							query.row.opponentPlayer,
+						)?.health,
+					).toBe(EthosLabCommon.health - 30 /* Anvil */)
+					expect(
+						game.components.find(
+							RowComponent,
+							query.row.index(1),
+							query.row.opponentPlayer,
+						)?.health,
+					).toBe(EthosLabCommon.health - 10 /* Anvil */)
```
Screaper91, line 112: Modify this test to check AFK Etho's health after playerTwo ends their turn
```suggestion
					yield* endTurn(game)
					expect(
						game.components.find(
							RowComponent,
							query.row.index(1),
							query.row.currentPlayer,
						)?.health,
					).toBe(EthosLabCommon.health - 10 /* Anvil only */)
```

Screaper91 (changes requested):
This looks good, I'll share additional `onTurnEnd` tests for other cards affected by these changes before this is merged.

common/status-effects/invisibility-potion.ts
Lunarmagpie: Good catch! I didn't realize it should last multiple turns.

Lunarmagpie:
> This looks good, I'll share additional `onTurnEnd` tests for other cards affected by these changes before this is merged.

I sent you a collaborator invite for my fork if you would like to write them to my branch.

PR #880, opened 2024-09-08
Improve Turtle shell behavior

Screaper91:
- Turtle Shell will no longer prevent damage from Poison and Burn
- Turtle Shell is no longer discarded when row is first activated on opponent's turn (Knockback)
- Fixes interactions with Emerald, rare Grian, Ladder, and Mending
_(Emerald and Grian can still re-attach Turtle Shell to an active hermit)_
- Adds visual status effect for when Turtle Shell is not protecting the active hermit
_(Turtle Shell was attached to that hermit after the hermit's first turn active)_

Other changes
- Fixes missing logs for changing to an new active row with no previous and changing to/from Armor Stand
- Fixes Mending apply log displaying invalid value
- Adds logs for single use cards that never created any


common/cards/alter-egos/effects/turtle-shell.ts
```diff
@@ -19,35 +53,158 @@ const TurtleShell: Attach = {
...
 		game: GameModel,
 		component: CardComponent,
 		observer: ObserverComponent,
 	) {
 		const {player} = component
 		let activated = false
+		let shouldDiscard = false
+
+		if (query.slot.active(game, component.slot)) {
+			// Only possible with Emerald or Grian's Borrow
```
Lunarmagpie, line 129: It would be better to give a more generic explanation here. It is too likely for comments that mention something specific to  become wrong in the future and confuse someone.
```
// Only possible with cards that move attachables. For example, Emerald or Grian's Borrow.
``

**Comment by Lunarmagpie** [comment]
I'm having a bit of trouble wrapping my head around the implementation so its gonna take a bit before I can give this a proper review.

**Comment by Screaper91** [comment]
> Is it possible you can add a test for using Turtle Shell with chest? I am worried that the current implementation will not handle that case properly.

Chest moves a card from the player's discard to their hand. LooseShellEffect and hook subscriptions are removed when Turtle Shell leaves the board or changes sides, so I'm not sure where Chest would create a problem case.

> I'm having a bit of trouble wrapping my head around the implementation so its gonna take a bit before I can give this a proper review.

I marked this ready for review because it was functional, not necessarily elegant, but it follows most players' behavior expectations.

File: common/cards/alter-egos/effects/turtle-shell.ts [hunk 2]
```diff
@@ -1,10 +1,44 @@
...
+})
+
+const updateRecord = (
+	game: GameModel,
+	player: PlayerEntity,
+	value: ActiveInfo,
+) => {
+	lastActiveHermit.get(game)[player] = value
+	console.info(
+		`${value.hermit.props.name} (${value.hermit.slot.inRow() && value.hermit.slot.row.index + 1}) is now "${value.stage}"`,
```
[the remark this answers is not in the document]
Screaper91: I'll add logs within `TurtleShell.onAttach` so that it only shares information about the hermit Turtle Shell is attached to.

common/status-effects/loose-shell.ts
```diff
@@ -0,0 +1,13 @@
+import {CardComponent} from '../components'
+import {StatusEffect, systemStatusEffect} from './status-effect'
+
+const LooseShellEffect: StatusEffect<CardComponent> = {
+	...systemStatusEffect,
+	icon: 'loose-shell',
+	id: 'loose-shell',
+	name: 'Loose Shell',
+	description:
+		'Turtle Shell will not protect this hermit until they first go AFK.',
```
Lunarmagpie, line 10:
```suggestion
		'Turtle Shell will not protect this hermit until the next time they become active.'
```

Lunarmagpie (changes requested):

Lunarmagpie:
I think the code will be a lot simpler to work with if you refactor to only use the state that the turtle shell is in instead of the state and whether the turtle shell is active and should be discarded.

Screaper91:
The only concern from using `onCreate` is that if that card needed to be introduced mid-game. If we were to add a single use card that added 3 random attach effects in your hand, it would need to call `onCreate` for every card in the available pool in case it was not already included in either player's deck.
```
const cardPool: Array<Card> = [TurtleShell, LightningRod, ThornsII, Wolf, DiamondArmor]
const AttachBoosterPack: SingleUse = {
  ...
  onCreate(game: GameModel, component: CardComponent) {
    cardPool.forEach((card: Card) => card.onCreate(game, component))
  }
  ...
}
```

Lunarmagpie:
> The only concern from using `onCreate` is that if that card needed to be introduced mid-game. If we were to add a single use card that added 3 random attach effects in your hand, it would need to call `onCreate` for every card in the available pool in case it was not already included in either player's deck.
> 
> ```
> const cardPool: Array<Card> = [TurtleShell, LightningRod, ThornsII, Wolf, DiamondArmor]
> const AttachBoosterPack: SingleUse = {
>   ...
>   onCreate(game: GameModel, component: CardComponent) {
>     cardPool.forEach((card: Card) => card.onCreate(game, component))
>   }
>   ...
> }
> ```

I am not very concerned about this at the moment. If it becomes an issue the card can always be rewritten.

Lunarmagpie:
Looks good. I just need @alicetouhou to look over all the battle logs to make sure they are good.

alicetouhou:
Status effect description looks good. However, the battle log descriptions are a bit overly complicated. They don't need to include reasoning; if the reasoning is unclear it needs to be in the card description, but I actually think it's fine in this case. They also shouldn't include when the protection goes away. There should be a log for when the protection starts, and for when it goes away.

Screaper91:
New logs should indicate when Turtle Shell starts protecting its hermit from opponent attacks and when it stops. It does not add a log to say its hermit is no longer protected if there is another log that indicates Turtle Shell was removed.

Lunarmagpie (approved):

common/cards/alter-egos/effects/turtle-shell.ts
```diff
@@ -1,10 +1,46 @@
...
+	| 'inactive'
+
+const lastActiveHermit = new GameValue<
+	Record<PlayerEntity, ActiveInfo | undefined>
+>(() => {
+	return {}
+})
+
+const startProtectionEntry = (hermit: CardComponent) =>
+	`$p${hermit.props.name}$ is now protected by their $e${TurtleShell.name}$`
```
Lunarmagpie, line 40:
```suggestion
	`$p${hermit.props.name}$ is now protected by $e${TurtleShell.name}$`
```

common/cards/alter-egos/effects/turtle-shell.ts
```diff
@@ -1,10 +1,46 @@
...
+const lastActiveHermit = new GameValue<
+	Record<PlayerEntity, ActiveInfo | undefined>
+>(() => {
+	return {}
+})
+
+const startProtectionEntry = (hermit: CardComponent) =>
+	`$p${hermit.props.name}$ is now protected by $e${TurtleShell.name}$`
+const endProtectionEntry = (hermit: CardComponent, rowIndex: number) =>
+	`$p${hermit.props.name}$ is no longer protected by $e${TurtleShell.name} (${rowIndex + 1})$`
```
Lunarmagpie, line 42:
```suggestion
	`$p${hermit.props.name}$ is no longer protected by $e${TurtleShell.name}$`
```

Screaper91: I included the Shell's row number here to avoid log ordering issues, for example a player changes between two Turtle Shells and either both hermits have the same name or `startProtectionEntry` comes before `endProtectionEntry` and creates ambiguous logs.

Lunarmagpie: I actually had to move it to after the hermits name, misunderstood what this was for.

Lunarmagpie:
I just did a lot of random testing with grains, chorus fruit, and ladder and could not find anything that didn't work as expected!

alicetouhou:
Who is this grain character?

Lunarmagpie:
I have updated the "lost protection" logs to use the hermit's row at @alicetouhou's request.

PR #886, opened 2024-09-11
Add type advantage tests

blockgoblin31:
Adds tests for type advantage.

Currently the tests with speedrunner attacking are failing as if the type advantage is not being applied, but I can't find a reason why. Edit: found the reason why

### DIFF
--- a/tests/game/type-advantage.test.ts
+++ b/tests/game/type-advantage.test.ts
@@ -1,7 +1,6 @@
 import {describe, expect, test} from '@jest/globals'
 import HotguyCommon from 'common/cards/alter-egos-iii/hermits/hotguy-common'
 import PoultrymanCommon from 'common/cards/alter-egos/hermits/poultryman-common'
-import {Hermit} from 'common/cards/base/types'
 import GoodTimesWithScarCommon from 'common/cards/default/hermits/goodtimeswithscar-common'
 import GrianCommon from 'common/cards/default/hermits/grian-common'
 import IJevinCommon from 'common/cards/default/hermits/ijevin-common'
@@ -14,20 +13,23 @@ import query from 'common/components/query'
 import {attack, endTurn, playCardFromHand, testGame} from './utils'
 
 let typeToHermitMap = {
-	'Prankster': PoultrymanCommon,
-	'PvP': XBCraftedCommon,
-	'Redstoner': MumboJumboCommon,
-	'Speedrunner': HotguyCommon,
-	'Farm': ImpulseSVCommon,
-	'Explorer': IJevinCommon,
-	'Builder': GrianCommon,
-	'Terraformer': GoodTimesWithScarCommon,
-	'Miner': TinFoilChefCommon,
+	Prankster: PoultrymanCommon,
+	PvP: XBCraftedCommon,
+	Redstoner: MumboJumboCommon,
+	Speedrunner: HotguyCommon,
+	Farm: ImpulseSVCommon,
+	Explorer: IJevinCommon,
+	Builder: GrianCommon,
+	Terraformer: GoodTimesWithScarCommon,
+	Miner: TinFoilChefCommon,
 }
 
 type HermitType = keyof typeof typeToHermitMap
 
-function createTypeAdvantageTest(attackingType: HermitType, defendingType: HermitType) {
+function createTypeAdvantageTest(
+	attackingType: HermitType,
+	defendingType: HermitType,
+) {
 	test(attackingType + ' v ' + defendingType, () => {
 		const attacker = typeToHermitMap[attackingType]
 		const defender = typeToHermitMap[defendingType]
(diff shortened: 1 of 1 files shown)


tests/game/type-advantage.test.ts
```diff
@@ -0,0 +1,58 @@
...
+                                query.row.currentPlayer,
+                                query.row.index(0)
+                            )?.health,
+                        ).toBe(defender.health - (attacker.primary.damage + 20))
+                    }
+                },
+                {startWithAllCards: true, noItemRequirements: true}
+            )
+        }
+    }
```
Lunarmagpie, line 57: It would be better to define this function outside of this scope.

Lunarmagpie (changes requested):

Lunarmagpie:
Please run the linter and formatter with `npm run format` and `npm run fix`

tests/game/type-advantage.test.ts
blockgoblin31: Yeah I had the same thought, I'm working on a slightly bigger refactor right now so itll just take type names as parameters. I can move the scope during that.

tests/game/type-advantage.test.ts
```diff
@@ -0,0 +1,72 @@
...
+import IJevinCommon from 'common/cards/default/hermits/ijevin-common'
+import ImpulseSVCommon from 'common/cards/default/hermits/impulsesv-common'
+import MumboJumboCommon from 'common/cards/default/hermits/mumbojumbo-common'
+import TinFoilChefCommon from 'common/cards/default/hermits/tinfoilchef-common'
+import XBCraftedCommon from 'common/cards/default/hermits/xbcrafted-common'
+import {RowComponent} from 'common/components'
+import query from 'common/components/query'
+import {attack, endTurn, playCardFromHand, testGame} from './utils'
+
+describe('Type advantage tests', () => {
```
Lunarmagpie, line 16: It's better to place this test at the bottom of the file. It's generally better to use things after they are defined even when it's not required.

tests/game/type-advantage.test.ts
```diff
@@ -0,0 +1,72 @@
...
+	['Prankster', PoultrymanCommon],
+	['PvP', XBCraftedCommon],
+	['Redstoner', MumboJumboCommon],
+	['Speedrunner', HotguyCommon],
+	['Farm', ImpulseSVCommon],
+	['Explorer', IJevinCommon],
+	['Builder', GrianCommon],
+	['Terraformer', GoodTimesWithScarCommon],
+	['Miner', TinFoilChefCommon],
+])
```
Lunarmagpie, line 26:
```suggestion
let typeToHermitMap: Map<String, Hermit> = {
	'Prankster': PoultrymanCommon,
	'PvP': XBCraftedCommon,
	'Redstoner': MumboJumboCommon,
	'Speedrunner': HotguyCommon,
	'Farm': ImpulseSVCommon,
	'Explorer': IJevinCommon,
	'Builder': GrianCommon,
	'Terraformer': GoodTimesWithScarCommon,
	'Miner': TinFoilChefCommon,
}
```

Lunarmagpie (changes requested):

tests/game/type-advantage.test.ts
blockgoblin31: It being var is just my bad, but typescript complains if I try to declare the map like a dictionary. The exact error is "Object literal may only specify known properties, and ''Prankster'' does not exist in type 'Map<String, Hermit>'."

Lunarmagpie: it should be `string` in lowercase I believe.

blockgoblin31: That gives a different error when I try to get from it(using typeToHermitMap[attackingType])
`Element implicitly has an 'any' type because expression of type 'string' can't be used to index type '{ Prankster: Hermit; PvP: Hermit; Redstoner: Hermit; Speedrunner: Hermit; Farm: Hermit; Explorer: Hermit; Builder: Hermit; Terraformer: Hermit; Miner: Hermit; }'.
  No index signature with a parameter of type 'string' was found on type '{ Prankster: Hermit; PvP: Hermit; Redstoner: Hermit; Speedrunner: Hermit; Farm: Hermit; Explorer: Hermit; Builder: Hermit; Terraformer: Hermit; Miner: Hermit; }'`

tests/game/type-advantage.test.ts
```diff
@@ -0,0 +1,72 @@
...
+
+function createTypeAdvantageTest(attackingType: String, defendingType: String) {
+	test(attackingType + ' v ' + defendingType, () => {
+		let attacker = typeToHermitMap.get(attackingType)!
+		let defender = typeToHermitMap.get(defendingType)!
```
Screaper91, line 31: This should work with Lunar's suggestion
```suggestion
type HermitType = keyof typeof typeToHermitMap

function createTypeAdvantageTest(attackingType: HermitType, defendingType: HermitType) {
	test(attackingType + ' v ' + defendingType, () => {
		const attacker = typeToHermitMap[attackingType]
		const defender = typeToHermitMap[defendingType]
```

Lunarmagpie (approved):

Lunarmagpie:
Thank you!

PR #887, opened 2024-09-12
Jopacity fixes

Screaper91:
- Fixes Jopacity dealing too much damage when opponent's Invisibility flips tails
- Fixes Puppetry/Role Play continuing to deal reduced damage until a primary is used after using Jopacity

### DIFF
--- a/common/models/attack-model.ts
+++ b/common/models/attack-model.ts
@@ -162,7 +162,7 @@ export class AttackModel {
 
 	// Setters / modifier methods
 
-	/** Increases the damage the attack does */
+	/** Increases the base damage the attack does */
 	public addDamage(source: AttackerEntity, amount: number) {
 		if (this.damageLocked) return this
 		this.damage += amount
@@ -172,7 +172,17 @@ export class AttackModel {
 		return this
 	}
 
-	/** Reduces the damage the attack does */
+	/** Reduces the base damage of the attack (before multiplication) */
+	public removeDamage(source: AttackerEntity, amount: number) {
+		if (this.damageLocked) return this
+		this.damage -= amount
+
+		this.addHistory(source, 'remove_damage', amount)
+
+		return this
+	}
+
+	/** Reduces the total damage the attack does */
 	public reduceDamage(source: AttackerEntity, amount: number) {
 		if (this.damageLocked) return this
 		this.damageReduction += amount
--- a/common/cards/alter-egos-iii/hermits/beetlejhost-rare.ts
+++ b/common/cards/alter-egos-iii/hermits/beetlejhost-rare.ts
@@ -6,7 +6,7 @@ import {
 import query from '../../../components/query'
 import {GameModel} from '../../../models/game-model'
 import ChromaKeyedEffect from '../../../status-effects/chroma-keyed'
-import {afterAttack} from '../../../types/priorities'
+import {afterAttack, beforeAttack} from '../../../types/priorities'
 import {hermit} from '../../base/defaults'
 import {Hermit} from '../../base/types'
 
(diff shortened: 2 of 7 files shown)


Lunarmagpie (changes requested):
I don't like that the chroma keyed logic has been moved to Beetlejhost. I think that effects should be able to be applied and work on any hermit to keep the code easy to follow.

Screaper91:
Chroma Keyed is single purposed for Beetlejhost's Jopacity, it would not make sense for it to be re-used for another decaying special attack because then swapping between the two attacks would not reset the counter. I could move the damage reduction logic back, but it would require adding something like `AttackModel.isOrigin(source: Card): boolean` to check that `BeetlejhostRare` created the attack

Screaper91:
> I think that effects should be able to be applied and work on any hermit to keep the code easy to follow.

The reviewed changes are tested to check that other hermits using Beetlejhost's Jopacity with mocked attacks still reduce their damage and remove Chroma Keyed at a consistent time.

Lunarmagpie:
> Chroma Keyed is single purposed for Beetlejhost's Jopacity, it would not make sense for it to be re-used for another decaying special attack because then swapping between the two attacks would not reset the counter. I could move the damage reduction logic back, but it would require adding something like `AttackModel.isOrigin(source: Card): boolean` to check that `BeetlejhostRare` created the attack

I see what you mean. I think I had a hard time articulating my thoughts for this.
When someone looks at chroma keyed, everything should be there. I find that splitting up parts of the effect between the hermit and the effect leads to an issue where you don't know where the stuff is happening as a new developer, and once you do know it makes it harder to iterate on the code because the logic is split up. I also feel like this implementation is working around the Zombie Cleo and Rendog bugs instead of dealing with the root of the issue.

I feel like mock cards should be changed a bit to use the status effects of the target hermit. I think that might be the right move here.

Lunarmagpie:
I'm going to write my own version of the Cleo/Ren fixes. We probably can compare them. I feel like I need to program it to understand the full scope of this issue.

Lunarmagpie:
Just finished looking into this myself. I don't have much to show yet, but I feel something like `Attack.isOrigin` is the right way to go about this.

Lunarmagpie:
Actually I think this PR is the best way to do this, I have come full circle lol

Lunarmagpie:
Can you elaborate on why we need both `removeDamage` and `reduceDamage`?

Screaper91:
- `removeDamage` is for attacks like Jopacity and World Eater that may deal less damage than `Hermit.secondary.damage`.
- `reduceDamage` is for Armor/Shield which reduce the total damage after damage multiplications.

Consecutive Jopacity vs Invisibility tails and Diamond Armor should be calculated as:
```
(
   Beetlejhost.secondary.damage /* Base Hermit damage */
   - 10 * ChromaKeyed.counter) /* Attack power reduction */
) * 2 /* Damage multiplier (e.g. Invisibility Tails) */
- 30 /* Total damage reduction (e.g. Diamond Armor) */
```
The hackier equivalent of `attack.removeDamage(component.entity, 10)` would be `attack.addDamage(component.entity, -10)`

Lunarmagpie:
I see. That makes sense.

Lunarmagpie (approved):

common/models/attack-model.ts
```diff
@@ -172,7 +172,17 @@ export class AttackModel {
 		return this
 	}
 
-	/** Reduces the damage the attack does */
+	/** Reduces the base damage of the attack (before multiplication) */
+	public removeDamage(source: AttackerEntity, amount: number) {
```
Lunarmagpie, line 176: I think this function should actually be called `subtractDamage` so it's clear it's the opposite of add. I think that would clear up the confusion I had a bit.

Lunarmagpie: Or maybe keeping this as `removeDamage` is fine, ill just go back and rename the other function since its kinda on the edge of what fits in this PR.

PR #889, opened 2024-09-12
Add tests for a couple different older bugs

blockgoblin31:
Tests added:
 - Add test for gaslight where afk hermit was attacked but took no damage https://github.com/hc-tcg/hc-tcg/pull/834
 - Add test for thorns with effect card https://github.com/hc-tcg/hc-tcg/pull/779
 - Add test for armor stand(Two bugfixes in 0.8.4)
 - Add test for anvil drop(Bugfixed in 0.8.3)

### DIFF
--- a/tests/game/skizzleman-rare.test.ts
+++ b/tests/game/skizzleman-rare.test.ts
@@ -1,12 +1,12 @@
 import {describe, expect, test} from '@jest/globals'
 import Anvil from 'common/cards/alter-egos/single-use/anvil'
+import IronArmor from 'common/cards/default/effects/iron-armor'
 import Thorns from 'common/cards/default/effects/thorns'
 import EthosLabCommon from 'common/cards/default/hermits/ethoslab-common'
 import SkizzlemanRare from 'common/cards/season-x/hermits/skizzleman-rare'
 import {RowComponent} from 'common/components'
 import query from 'common/components/query'
 import {attack, endTurn, playCardFromHand, testGame} from './utils'
-import IronArmor from 'common/cards/default/effects/iron-armor'
 
 describe('Test Skizzleman Rare', () => {
 	test('Gaslight works as intended', () => {
@@ -80,7 +80,7 @@ describe('Test Skizzleman Rare', () => {
 			{startWithAllCards: true, noItemRequirements: true},
 		)
 	})
-	test('Gaslight doesn\'t trigger if the hermit takes no damage', () => {
+	test("Gaslight doesn't trigger if the hermit takes no damage", () => {
 		testGame(
 			{
 				playerOneDeck: [EthosLabCommon, EthosLabCommon, IronArmor],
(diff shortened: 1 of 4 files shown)


Lunarmagpie:
Thanks for the work! The armor stand tests will almost certainly be useful!

tests/game/skizzleman-rare.test.ts
```diff
@@ -79,4 +80,31 @@ describe('Test Skizzleman Rare', () => {
...
+				playerOneDeck: [EthosLabCommon, EthosLabCommon, IronArmor],
+				playerTwoDeck: [SkizzlemanRare, Anvil],
+				saga: function* (game) {
+					yield* playCardFromHand(game, EthosLabCommon, 'hermit', 0)
+					yield* playCardFromHand(game, EthosLabCommon, 'hermit', 1)
+					yield* playCardFromHand(game, IronArmor, 'attach', 1)
+					yield* endTurn(game)
+
+					yield* playCardFromHand(game, SkizzlemanRare, 'hermit', 0)
+					//Use anvil to trigger attack on the afk hermit that does no damage
```
Lunarmagpie, line 95:
```suggestion
					// Use anvil to trigger attack the afk hermit for zero damage
```

Lunarmagpie (changes requested):
Right now the the thorns test only tests the Thorns I implementation and not Thorns II or Thorns III. I think theres two ways to go around it.
- Option 1: Extend the tests to test all of Thorns, Thorns II, and Thorns III.
- Option 2: Update Thorns so that they all have the same implementation. I was thinking there should be a function called `createThorns` or something that could return the thorns implementation based on some values so we don't need to duplicate code for all three thorns.

I would prefer option two, but if you want to go that route it would be best to remove the thorns test from the current PR and make a new one for just making thorns all depend on the same implementation.

blockgoblin31:
> Right now the the thorns test only tests the Thorns I implementation and not Thorns II or Thorns III. I think theres two ways to go around it.
> 
> * Option 1: Extend the tests to test all of Thorns, Thorns II, and Thorns III.
> * Option 2: Update Thorns so that they all have the same implementation. I was thinking there should be a function called `createThorns` or something that could return the thorns implementation based on some values so we don't need to duplicate code for all three thorns.
> 
> I would prefer option two, but if you want to go that route it would be best to remove the thorns test from the current PR and make a new one for just making thorns all depend on the same implementation.

I just assumed they all used the same code, probably should have checked. I’ll look into doing a refactor.

blockgoblin31:
I feel like refactoring the thorns is a bit of a bigger task because there's a bunch of other effects that really should be refactored too for consistency, so for now I'll add tests for each one separately

Lunarmagpie (approved):
LGTM

blockgoblin31:
I’ve realized there’s a potential bug with the skizz test, I’ll fix it this afternoon

PR #893, opened 2024-09-13
Fix potential bug with new rare skizz test

blockgoblin31:
I was worried that my new test could potentially return a false negative if the iron armor was bugged and blocked the whole 20 damage from gaslight, even after the ten from anvil. Gold armor should never do that if there isn't some very obvious bug.


Lunarmagpie (approved):
Sounds reasonable.

PR #894, opened 2024-09-14
Fix Failed Beetlejhost Tests

Lunarmagpie:


blockgoblin31:
Am I reading the new test wrong or is it checking the opposite of the description? As far as k can tell it’s expecting the effect to go away after doing nothing for a turn.

Lunarmagpie:
> Am I reading the new test wrong or is it checking the opposite of the description? As far as k can tell it’s expecting the effect to go away after doing nothing for a turn.

Its looking for there to be "no status effect that does not have a target". When status effects are removed from the board we set the target to `null`. So its looking to see that there is not a ChromaKeyed effect that has been removed from the board.

Lunarmagpie:
Im gonna skip waiting to see if anyone wants to review for this PR because the project is in a state where people can't work on it from the failed test.

PR #896, opened 2024-09-14
Remove random file

Lunarmagpie:
Not sure how this happened, theres three authors on that PR


Rvtar:
It went out of use when you did the refactor.

PR #898, opened 2024-09-15
Add subfolders to tests/game

blockgoblin31:
At this point there are a lot of tests, and I figured we should probably start adding subfolders for tests for specific cards.


Lunarmagpie:
I prefer to keep files in one directory but I am not the only person who works on this project so I would lik to hear what other contributors think

Screaper91:
I think having these tests in subfolders is a good idea to help organize tests as we continue to add more "testing suite" files for hermits, effects, and general game behavior.

Lunarmagpie:
I decided to merge this. I think keeping general game behavior away from other tests helps see what is tested at a glance.

PR #902, opened 2024-09-18
Add Evil X Boss

Lunarmagpie:
This pull request is for tracking progress on the Evil X boss. Supersedes #470.


Lunarmagpie:
I plan to investigate the sound effect code for this.

alicetouhou:
Just a note, testing evil X and he says his attack move sound effect like 5+ times. The damage also sometimes seemed higher than it was supposed to be. I also didn't lose even though Evil X gave his speech, not sure why because it does look like it's tested.

Screaper91:
> Just a note, testing evil X and he says his attack move sound effect like 5+ times. The damage also sometimes seemed higher than it was supposed to be. I also didn't lose even though Evil X gave his speech, not sure why because it does look like it's tested.

Can you share more context, I haven't been able to recreate these issues. I've been testing with `npm:build` & `npm:start` locally with `startWithAllCards: true` and a mono-type prankster deck.

Lunarmagpie:
I got the UI basically where I want it to be besides for a bug where entering an invalid code refreshes the screen.

alicetouhou:
> Can you share more context, I haven't been able to recreate these issues. I've been testing with `npm:build` & `npm:start` locally with `startWithAllCards: true` and a mono-type prankster deck.

I've been testing with `npm run client:dev` and `server:dev`. Did not build after checking out this branch however. My debug config is default (same as the example). If you can't replicate maybe it's just a weird issue because i didn't build.

alicetouhou:
Also, for "higher than it's supposed to be", his dialogue would say 50 but he'd deal 100 ect.Forgot to check what the number was in the battle log, sorry. I can't recall this happening in the opposite direction but it's 100% possible I missed it or forgot.

alicetouhou:
Was unable to replicate the problems I faced

Lunarmagpie:
> Was unable to replicate the problems I faced

In that case I think we are really really close to merging

alicetouhou:
> In that case I think we are really really close to merging

Yes, I'm not sure what else needs to be done 👍

Lunarmagpie:
Yep! I think this is done! 
Theres some tiny things I will probably make PRs for but they are issues on the top of the layer cake so  there is nothing to block anymore!

Screaper91 (approved):

Lunarmagpie (commented):
Approved (Why can't github let me approve my own PR??!!)

Lunarmagpie:
Merged!

PR #904, opened 2024-09-19
Fix Turtle Shell interaction with Extra Flee

Screaper91:
- Fixes Turtle Shell not discarding itself when activated by Extra Flee knock-out


Lunarmagpie (approved):
We have so many tests there that I trust it works.

PR #907, opened 2024-09-21
API Rewrite

Lunarmagpie:
This PR removes the API. I reduced it down to two endpoints for the time being:
- `/api/cards` - Get information about all the cards in the game
- `/api/deck/{DECK_CODE}` - Get the cards in a deck.

These two methods should make it far easier for anyone to interface with the game.


alicetouhou:
Personally I think it's ok if we need to have only these 2 endpoints for a few weeks. Showing amount of games seems to currently be broken on the bot anyway. and it's worth it if we can eventually add more features

Lunarmagpie:
> Personally I think it's ok if we need to have only these 2 endpoints for a few weeks. Showing amount of games seems to currently be broken on the bot anyway. and it's worth it if we can eventually add more features

Yeah, I think you are right. This PR is done in that case.

PR #908, opened 2024-09-22
Switch over to storing data using Postgres SQL

alicetouhou:
This PR removes firebase and replaces it with a PSQL database.
- we will have a complete history of games, which will allow for much more detailed stats in the future
  - in addition, we don't need to rely on firebase anymore, and i personally believe my system is a lot more robust and better suited for this project
- players will be able to sync all their data across devices
- decks are no longer restricted to only being able to use type icons
- codes to share decks are now only 7 hex digits, we don't need to encode them into b64 anymore
- framework for storing game replays and adding achievements in the future is in place

Completed:
- Creating decks
- Selecting decks
- Exporting
- Deleting decks
- Adding games once they are complete
- Updating deck
- Tags
- Deck Validation
- Active decks
- Copying decks
- Stats
- Syncing accounts
- Transfer local data on first login
- Starter deck
- Importing

Testing is completed to a degree I feel comfortable with, but other people are always welcome to add tests. There's also no API tests, because the API isn't set up yet.

TBD:
- Visual improvements
- API

IMPORTANT NOTE: Removes legacy deck support. However, because of the server transfer, nobody should have these anymore anyway.


alicetouhou:
I'm having weird issues but hoping I can fix this after my midterm

alicetouhou:
This is not ready to merge, but I definitely need reviews at this point. I would appreciate if the other developers could run this version for themselves to make sure it works.

alicetouhou:
Ok I'm really ready for review now!

Lunarmagpie:
Sorry, im gonna be a bit before I work on the e2e tests. I want to finish my keyboard and I have to finish my math homework due tonight first.

PR #910, opened 2024-09-23
Move Socket Into Redux Store

Lunarmagpie:
The socket being a global variable was causing issues when I was working on testing the frontend. Additionally, its just good practice to not have globals hanging around.

This seems to work, but I am a little worried that I made a mistake here so it would be awesome if I can have other people check this!


alicetouhou:
lgtm

PR #913, opened 2024-09-24
Add Playwright

Lunarmagpie:
This PR adds playwright so we can improve over testing coverage. I would like to add one real effective test before this is merged.


alicetouhou:
lgtm

PR #914, opened 2024-09-24
Add Snapshot Tests For Card Rendering

Lunarmagpie:
Note that this test will require the snapshots to be regenered if certain token costs are changed. Although its a little annoying, its better than hoping that they are drawn correctly.


alicetouhou:
I'm a bit confused, what is the purpose of this test?

Lunarmagpie:
> I'm a bit confused, what is the purpose of this test?

Remember how I broke items so they displayed as the type every time? This test ensures that cards always display as a correct version which would prevent that issue in the future.

alicetouhou:
Ah ok. Looks good but why is there a show cost prop now?

Lunarmagpie:
> Ah ok. Looks good but why is there a show cost prop now?

It removes a dependency on the game object global existing. This made is extremely difficult to test so I had to change this.

alicetouhou:
Ok. Still a bit confused but I can't really look closer bc no wifi. Not easy to review right now

PR #918, opened 2024-09-25
Consolidate hooks

Screaper91:
- Moves attack hooks and `freezeSlots` to GameModel
- Removes `beforeDefence` and `afterDefence` hooks
- Adds type `ReadonlyAttackModel` to prevent modifying attacks after execution
- Fixes issues with reviving rows by adding a dedicated hook to the attack cycle
- Adds explicit typing to calling hooks

### DIFF
--- a/common/cards/alter-egos-ii/hermits/boomerbdubs-rare.ts
+++ b/common/cards/alter-egos-ii/hermits/boomerbdubs-rare.ts
@@ -155,9 +155,10 @@ const BoomerBdubsRare: Hermit = {
 		)
 
 		observer.subscribeWithPriority(
-			player.hooks.beforeAttack,
+			game.globalHooks.beforeAttack,
 			beforeAttack.MODIFY_DAMAGE,
 			(attack) => {
+				if (attack.player.entity !== player.entity) return
 				if (!attack.isAttacker(component.entity) || attack.type !== 'secondary')
 					return
 				if (flippedTails === true) {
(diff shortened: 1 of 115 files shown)


Lunarmagpie:
First, `game.globalHooks` should be renamed to `game.hooks`. I don't think the longer name is worth it since this is used so many times and its pretty clear already.
Second, I'm not sure this is the best approach. Since every use of these hooks have the if statement, I feel like its more effective to keep it scoped to the correct player.

alicetouhou:
> Since every use of these hooks have the if statement, I feel like its more effective to keep it scoped to the correct player.
I agree with that. I find this a bit more confusing to read. Would probably get used to it, though.

Lunarmagpie:
Code looks good overall, need some time to think about if this is the best approach though.

alicetouhou:
> > Since every use of these hooks have the if statement, I feel like its more effective to keep it scoped to the correct player.
> 
> I agree with that. I find this a bit more confusing to read. Would probably get used to it, though.

Thought about it yesterday and actually I have no issue with it. Sure it's different but that doesn't mean that's bad. The only real issue I can think of is it may sometimes add another indentation level but that case is totally rare, plus it simplifies other cases too. gonna go ahead and approve this

alicetouhou (approved):

Lunarmagpie:
I don't want to block this since everyone prefers it so we should merge.
I just want to add a test to verify armor stand freezes the correct slots.

Lunarmagpie:
I just remembered that this change also fixes a lot of difficulties with trying to get before attack and after attack to trigger in the right priority. It's definitely the right choice to merge this.

PR #921, opened 2024-09-26
Fix interactions with ultra-rare TFC and rare Worm Man

Screaper91:
**Worm Man's Total Anonymity**
- Fixed not flipping cards when mocked by Puppetry/Role Play
- Fixed being able to place multiple hermits after Worm Man is knocked out
- Fixed logs revealing cards placed face-down before they are turned face-up

**TFC's Take It Easy**
- Fixed being able to discard twice from the same hermit when mocked by Puppetry/Role Play
- Fixed discarding twice from a hermit after they are moved by Ladder
- Fixed not being able to discard if TFC already discarded from a previous hermit in that row


alicetouhou:
Looks good, although since this would probably cause conflicts with 918 do you want us to merge it first?

Screaper91:
Sounds good, there is only one merge conflict and I'll make sure these added tests pass with 918

PR #922, opened 2024-09-26
Fix playing milk/water bucket as single use

Screaper91:
Fixes Milk/Water Bucket only being playable as a single use depending on other cards in hand


alicetouhou:
Thank you looks good!

PR #925, opened 2024-09-28
Move Around Buttons On Title Screen

Lunarmagpie:
image


client/src/logic/messages.ts
Screaper91, line 1: `MATCHMAKING_PRIVATE_GAME_JOIN` and `MATCHMAKING_PRIVATE_GAME_CREATE` are still in the `type Messages` array

client/src/logic/matchmaking/matchmaking-reducer.ts
Screaper91, line 1: the case `MATCHMAKING_PRIVATE_GAME_CREATE` needs to be removed

client/src/app/match-making/match-making.tsx
```diff
@@ -35,6 +37,46 @@ function MatchMaking() {
...
+				returnText="Settings"
+				className={classNames(css.privateLobby)}
+			>
+				<div className={css.privateLobbyLeft}>
+					<p>Opponent Code</p>
+					<div className={css.code} onClick={handleCodeClick}>
+						{gameCode}
+					</div>
+					<p>Spectator Code</p>
+					<div className={css.code} onClick={handleCodeClick}>
```
Screaper91, line 54: This copies the opponent code, not the spectator code.

Screaper91 (changes requested):
There are a couple build errors that need addressing. Otherwise my only complaints are the the right side of private lobby UI should be neater and the code `div`'s need the `<CopyIcon />` svg.

client/src/app/match-making/match-making.tsx
Lunarmagpie: This has been fixed on master but needs to be merged.

Lunarmagpie:
The UI isn't done but I fixed the build errors. It might be best to merge this in now since the UI changes are far enough along that there isn't serious issues.

Screaper91 (approved):
Sounds good to merge functionality now and fix the UI later

PR #929, opened 2024-09-30
Rework Evil X To Allow For Sound Effects Without Side Effects

Lunarmagpie:
This PR makes a wait action and allows sounds to pushed to the game state. This makes it possible for ai turn functions to always be completely pure.

### DIFF
--- a/common/socket-messages/server-messages.ts
+++ b/common/socket-messages/server-messages.ts
@@ -41,7 +41,6 @@ export const serverMessages = messages({
 	GAME_OVER_STAT: null,
 	GAME_STATE: null,
 	CHAT_UPDATE: null,
-	VOICE_ANNOUNCE: null,
 })
 
 export type ServerMessages = [
--- a/common/socket-messages/server-messages.ts
+++ b/common/socket-messages/server-messages.ts
@@ -97,7 +97,6 @@ export type ServerMessages = [
 	},
 	{type: typeof serverMessages.GAME_STATE; localGameState: LocalGameState},
 	{type: typeof serverMessages.CHAT_UPDATE; messages: Array<ChatMessage>},
-	{type: typeof serverMessages.VOICE_ANNOUNCE; lines: Array<string>},
 ]
 
 export type ServerMessage = Message<ServerMessages>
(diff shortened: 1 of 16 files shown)


server/src/routines/turn-actions.ts
```diff
@@ -375,3 +376,9 @@ export function* pickRequestSaga(
 
 	return
 }
+
+export function* delaySaga(game: GameModel, delayMs: number) {
+	if (game.viewers.length !== 0) {
+		yield* delay(delayMs)
+	}
+}
```
Screaper91, line 384: This delays sending the game state until after the time elapses, it needs to begin the voice queue before waiting so it can finish by the time the next action is provided

Screaper91 (changes requested):
There is also a build error with `client/src/logic/game/tasks/action-modals-saga.ts`.

Otherwise I approve of this being refined then merged

common/socket-messages/server-messages.ts
```diff
@@ -41,7 +41,6 @@ export const serverMessages = messages({
 	GAME_OVER_STAT: null,
 	GAME_STATE: null,
 	CHAT_UPDATE: null,
-	VOICE_ANNOUNCE: null,
```
Screaper91, line 44: This was not deleted in the type ServerMessages

Lunarmagpie:
> There is also a build error with `client/src/logic/game/tasks/action-modals-saga.ts`.
> 
> Otherwise I approve of this being refined then merged

Ah my bad!

server/src/routines/turn-actions.ts
Lunarmagpie: I added the `sendGame` function before the `delaySaga`. This should be okay, as during tests the send function will do nothing because there is no viewers.

server/src/routines/game.ts
```diff
@@ -364,6 +367,7 @@ function* turnActionSaga(
 				'PICK_REQUEST',
 				'MODAL_REQUEST',
 				'END_TURN',
+				'DELAY',
```
Screaper91, line 370: This should be removed, it cancels `DELAY` actions unless it is in `availableActions` and does the next action immediately

Lunarmagpie: Fixed

Screaper91 (approved):
LGTM

PR #930, opened 2024-09-30
Rework End Turn Modal To Add Tests

Lunarmagpie:
In the the past, there have been several instances of the end turn modal displaying when it should not, or not displaying when it should. Overall it has been very broken.
This PR adds snapshot tests for it.


Lunarmagpie:
Unfortunately this is blocked until radix morals are removed

PR #937, opened 2024-10-01
Fix swapping after using an attack

Screaper91:
- Fixes Let's Go not allowing to switch hermits when mocked by Puppetry/Role Play
- Fixes Chorus Fruit allowing to switch hermits after attacking with Shreep
- Fixes Chorus Fruit being unusable if oppenent flipped 2 heads for Betrayed


Lunarmagpie (approved):
LGTM

Lunarmagpie:
I feel like blocked actions are not powerful enough, then we have frozen slots that do like the same thing, it would be good to think about if theres a better way.

Issue #600, opened 2024-07-08
Redo Blocked an Available Actions


Lunarmagpie:
- I feel available actions should be removed because they do the same thing as blocked actions but are too confusing.
- Blocked actions should be tied to what hermit its blocked for. This would allow JoeHills  & Cleo to work together properly for example. At the moment using Timeskip with Cleo's puppetry will block Cleo's puppetry.

Benji42:
I think this system seems a lot better, it keeps the benefits of the old system while allowing for more functionality.

Sense101:
If you mean the available actions hooks, hard agree. Was always planned

Lunarmagpie:
> If you mean the available actions hooks, hard agree. Was always planned

That will be part of it, yes. The main goal is to scope blocked actions to a specific card or slot though.

Sense101:
Pretty sure blocked actions already works with ids, but it's really finicky

Lunarmagpie:
It turns out the available actions hook is pretty useful. When we mimick attacks with Rendog or Zombiecleo it is helpful to be able to check specifically if the mimicked card wants to disable it's ability so we don't need to show it as an option. For this reason I think we should use the hook for blocking actions whenever possible.

PR #950, opened 2024-10-05
Fix consecutive Time Skip

Screaper91:
- Fixes mocking Time Skip after flipping heads
- Fixes actions being blocked/available at the wrong time due to knock-outs
- Fixes Role Play disabling all secondary attacks if opponent only has Armor Stand


Screaper91:
I know this does not neaten the blocked actions system for #600 but it does add tests that will hopefully prevent bugs with the next refactor

Lunarmagpie:
I just have a question about this interaction:
- Is the joehills time skip supposed to additionally prevent clock from being used, and vise versa?@alicetouhou

Lunarmagpie:
> I know this does not neaten the blocked actions system for #600 but it does add tests that will hopefully prevent bugs with the next refactor

Np, it's fine to work with this current system until someone has time to investigate and redo it

Lunarmagpie (approved):
Everything that made me question "why is that like that" ended up being because of a limitation with blocked actions :(. LGTM. Definitely should investigate redoing those.

alicetouhou:
> I just have a question about this interaction:
> 
>     * Is the joehills time skip supposed to additionally prevent clock from being used, and vise versa?@alicetouhou

correct

PR #951, opened 2024-10-06
Add Image Optimizer For Production Builds

Lunarmagpie:
This reduces the size of images by 69% according to the CI. It will make production builds take another minute but I think that is okay because the speed gains from this will be a lot since the application is so image heavy.


alicetouhou:
lgtm

PR #955, opened 2024-10-07
Fix guaranteed hermit reducing chance of drawing second

Screaper91:
If starting hand does not start with a hermit, reshuffles the deck pile until there is at least one. Fixes always swapping the first hermit in deck, which reduced the odds of drawing a second hermit.

### DIFF
--- a/common/cards/default/hermits/rendog-rare.ts
+++ b/common/cards/default/hermits/rendog-rare.ts
@@ -90,8 +90,6 @@ const RendogRare: Hermit = {
 					)
 					.apply(component.entity)
 			observer.subscribe(player.hooks.onTurnStart, () => {
-				// Ren having the "Blocked Action" icon on turn one looks weird
-				if (game.state.turn.turnNumber <= 1) return
 				if (!game.components.exists(SlotComponent, pickCondition))
 					game.components
 						.new(
--- a/common/cards/default/hermits/rendog-rare.ts
+++ b/common/cards/default/hermits/rendog-rare.ts
@@ -90,6 +90,8 @@ const RendogRare: Hermit = {
 					)
 					.apply(component.entity)
 			observer.subscribe(player.hooks.onTurnStart, () => {
+				// Ren having the "Blocked Action" icon on turn one looks weird
+				if (game.state.turn.turnNumber <= 1) return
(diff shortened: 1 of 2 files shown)


Sense101:
I believe this will increase the chances of drawing more than one hermit past normal random chance, I don't think this is a good idea

Screaper91:
**Example:**
Starting Hand size = 1
Deck size = 4
Hermit count = 2

```
HHEE 1/6 + 1/12 + 1/12 = 33%
HEHE 1/6 + 1/12 + 1/12 = 33%
HEEH 1/6 + 1/12 + 1/12 = 33%
EHHE -> 50% HEHE, 50% HHEE
EHEH -> 50% HEEH, 50% HHEE
EEHH -> 50% HEEH, 50% HEHE
```

Sense101:
Does this scale with a higher starting hand size though? I'm imagining a roll of no hermits, so we pick a random hermit, then a roll  where maybe the second and third cards are hermits, resulting in us replacing the first card with our random hermit and therefore having more hermits picked than there would normally be.

Am I missing something?

Screaper91:
Starting hand size = 2
Deck size = 4
Hermit count = 2
```
HHEE 1/6 = 17%
HEHE 1/6 + 1/24 = 21%
HEEH 1/6 + 1/24 = 21%
EHHE 1/6 + 1/24 = 21%
EHEH 1/6 + 1/24 = 21%
EEHH -> 25% each for HEHE, HEEH, EHHE, EHEH
```
@Lunarmagpie is there any better way to randomize the order of `DeckSlotComponent` to shuffle repeatedly

Lunarmagpie:
> I believe this will increase the chances of drawing more than one hermit past normal random chance, I don't think this is a good idea

This isn't relevant. This is how you deal with bad hands in analog card games.

Lunarmagpie:
It would be better to first create all the card components and sort them in an array. Then take that array and add it to the deck or hand as required.

Lunarmagpie (approved):
LGTM

Lunarmagpie:
> is there any better way to randomize the order of DeckSlotComponent to shuffle repeatedly

Maybe but I don't think its worth bothering. Theres basically a 0% chance we'll ever need to shuffle more than like 3 times.

PR #956, opened 2024-10-08
Move Cards To Higher Directory and Combine Similar Cards

Lunarmagpie:
This PR moves many cards to folders that don't depend on an expansion. This makes it easier to find a specific card and additionally lets us place card level versions in the same file without it being confusing.


Lunarmagpie:
I can go through and fix any issues this causes with the AoTCG branch.

Screaper91 (commented):
Looks okay to me, it puts a lot of hermit files together but it does make it easier to find the alter ego hermits

Lunarmagpie:
I'm going to wait to merge this until the AoTCG PR is merged.

alicetouhou (approved):

PR #957, opened 2024-10-08
Use Fisher Yates Shuffle

Lunarmagpie:
The previous shuffle algorithm was not fair. This fixes the issue by replacing the shuffling algorithm with fisher-yates which is fair.


common/utils/state-gen.ts
```diff
@@ -110,20 +111,16 @@ function setupEcsForPlayer(
...
 			!cards.slice(0, amountOfStartingCards).some((card) => card.isHermit())
 		) {
-			cards
-				.sort(() => Math.random() - 0.5)
-				.forEach((card, i) => {
-					if (card.slot.inDeck()) card.slot.order = i
-				})
+			fisherYatesShuffle(cards).forEach((card, i) => {
+				if (card.slot.inDeck()) card.slot.order = i
+			})
```
Screaper91, line 123: Can you rewrite this to use a `do...while` loop instead of repeating the same shuffle logic

Lunarmagpie: No, I do not like do whiles.

PR #959, opened 2024-10-09
Original weakness

Rvtar:
Somehow a can of worms since it's released.


Rvtar:
At this point, I haven't tested anything and there are still red underlines in my code.

alicetouhou:
I would recommend making a new branch and restating

Rvtar:
> I would recommend making a new branch and restating

You mean restart?

alicetouhou:
yes

Rvtar:
I have restarted a few times now. Should I still make a new branch for commit history?

Rvtar:
Now this should work properly once #967 is merged.

alicetouhou:
Your PR has a change in the debug config example file. Please make sure this file is the same as master.
In addition, you need to run linters and make sure tests still pass.

common/status-effects/weakness.ts
```diff
@@ -1,29 +1,49 @@
...
 	...statusEffect,
 	id: 'weakness',
 	icon: 'weakness',
 	name: 'Weakness',
-	description: "This Hermit is weak to the opponent's active Hermit's type.",
+	description:
+		'[weakType] is weak to [strongType] for the duration fo this counter.',
 	counter: 3,
 	counterType: 'turns',
 	onApply(
```
Screaper91, line 20: Do not apply status if the opponent's active hermit is Armor Stand. Your implementation will cause it to last forever and show the template description
```suggestion
	counterType: 'turns',
	applyCondition: (_game, value) =>
		value instanceof PlayerComponent && !!value.getActiveHermit()?.isHermit(),
	onApply(
```

common/status-effects/weakness.ts
```diff
@@ -1,29 +1,49 @@
...
-		observer.subscribe(player.hooks.onTurnStart, () => {
+		const playerActive = player.getActiveHermit()
+		const opponentActive = opponentPlayer.getActiveHermit()
+
+		if (!playerActive?.isHermit() || !opponentActive?.isHermit()) return
+
+		const weakType = playerActive.props.type
+		const strongType = opponentActive.props.type
+		function capitalize(s: string) {
+			return s && String(s[0]).toUpperCase() + String(s).slice(1)
```
Lunarmagpie, line 37: The `String` function call here doesn't do anything

Lunarmagpie (changes requested):

Lunarmagpie:
This just needs tests before it's merged I think

Rvtar:
What is there to test?

Rvtar:
What is there to test?

Lunarmagpie:
Test that both the AFK and Active hermit of the same type take extra damage, AFK hermits of the wrong type do not take damage, and that the effect times out properly

tests/unit/game/effects/weakness.test.ts
```diff
@@ -0,0 +1,320 @@
...
+					yield* playCardFromHand(game, SmallishbeansRare, 'hermit', 0)
+					yield* playCardFromHand(game, SmallishbeansRare, 'hermit', 1)
+					yield* playCardFromHand(game, SmallishbeansRare, 'hermit', 2)
+					yield* attack(game, 'secondary')
+					yield* endTurn(game)
+
+					yield* endTurn(game)
+
+					yield* playCardFromHand(game, PotionOfWeakness, 'single_use')
+					yield* attack(game, 'secondary')
```
Screaper91, line 278: Potion of Weakness must be manually applied when played because it shows the confirmation modal.
```suggestion
					yield* playCardFromHand(game, PotionOfWeakness, 'single_use')
					yield* applyEffect(game)
					yield* attack(game, 'secondary')
```

tests/unit/game/effects/weakness.test.ts
```diff
@@ -0,0 +1,320 @@
...
+
+					expect(
+						game.components.find(
+							RowComponent,
+							query.row.currentPlayer,
+							query.row.index(2),
+						)?.health,
+					).toBe(
+						SmallishbeansRare.health - SmallishbeansRare.primary.damage - 20,
+					)
```
Screaper91, line 314: These checks fail because the starting hermit health does not match up and is swapped

tests/unit/game/effects/weakness.test.ts
```diff
@@ -0,0 +1,320 @@
...
+					yield* playCardFromHand(game, PotionOfWeakness, 'single_use')
+					yield* attack(game, 'secondary')
+					yield* endTurn(game)
+
+					yield* pick(
+						game,
+						query.slot.currentPlayer,
+						query.slot.hermit,
+						query.slot.rowIndex(1),
+					)
```
Screaper91, line 286: This is a regular change active row action, not from using chorus fruit
```suggestion
					yield* changeActiveHermit(game, 1)
```

Screaper91 (commented):
Here are the issues causing `weakness.test.ts` to fail when run.

tests/unit/game/effects/weakness.test.ts
```diff
@@ -0,0 +1,321 @@
...
+					yield* endTurn(game)
+
+					yield* endTurn(game)
+
+					yield* playCardFromHand(game, PotionOfWeakness, 'single_use')
+					yield* applyEffect(game)
+					yield* attack(game, 'secondary')
+					yield* endTurn(game)
+
+					yield* changeActiveHermit(game, 1)
```
Screaper91, line 289: Since this is supposed to check Weakness still works after a knock-out, you need one more AFK Joel to deal enough damage. Add this line to verify a knock-out did occur
```suggestion
					expect(game.currentPlayer.activeRow).toBe(null)
					yield* changeActiveHermit(game, 1)
```

Rvtar: That damage was covered by weakness. Though that's definitely a useful addition.

Screaper91: Or apply weakness the first time you use the secondary since it is still 10hp short of a KO

Lunarmagpie:
The names of the tests are a bit unclear to me. The name of every single test should include the expected result. For example instead of "Weakness Damage Wrong Type Exchange" you could do "Attacking a type that is not affected by weakness does not take extra damage".

tests/unit/game/effects/weakness.test.ts
Rvtar: I forgot I used Impy for that part and my brain switched back to Joel for the rest of the construction lol.

Rvtar:
Renamed.

alicetouhou:
This is looking good to me 😀
one thing, j think "duration of this effect" would sound a bit better than "duration of this counter", but it's not that important so I'll approve you

alicetouhou (approved):

Lunarmagpie (approved):
LGTM

Screaper91 (approved):

PR #960, opened 2024-10-10
[WIP] Run Game On Client

Lunarmagpie:
This PR refactors the game to run the game on the client along with the server. This is not close to ready to review but I wanted to show I am working on this.

Adds tests to verify the game ended with the correct winner and tests that the end game modal displays correctly when given the game outcome and player information.

## Todo List
- [ ] Write test for reconnects during a game 
- [ ] Write tests for desync between client and server (entire game should be recalculated)
- [x] Fix a bug where coin flips force a reconnect
- [ ] Redo how 'DELAY' turn action works. For example there is an issue where forfeiting doesn't work while the delay is happening.
- [ ] Make sure clients can use actions when their timer reads 0 seconds left, ie, the action isn't thrown away once it reaches the server.
- [x] Multiple games can't run at the same time 
- [ ] Redo how sound effects work during boss fights   
- [x] Verify boss game reconnects do not have unnecessary delays.


alicetouhou:
If we want replays, this is needed before we merge the database things. Otherwise, we don't need this to be done beforehand

alicetouhou:
Scary merge conflicts...

Lunarmagpie:
I am going to redo this PR

PR #962, opened 2024-10-21
Fix minor issues with tag system

alicetouhou:
I'll be honest, this is not my best code. But HTML behavior is extremely strange sometimes, and at least this definitely works.

- Deck List updates properly when duplicating, importing, or deleting decks
- Tag filter works properly when loading into the screen
- Tag color doesn't change back to white after creating a tag. This was intended, but I'm not really sure why I decided to do this. Anyway, the old way I implemented it was wrong anyway so I fixed it.
- The deck screen ALWAYS scrolls to the correct position, so the deck you selected is always visible. New feature, but this fixes some issues too.

Would appreciate a review. This is very spaghetti

### DIFF
--- a/client/src/app/deck/deck-edit.tsx
+++ b/client/src/app/deck/deck-edit.tsx
@@ -128,7 +128,7 @@ const addTag = (
 	if (tags.length >= 3) return
 	if (tag.name.length === 0) return
 	setTags([...tags, tag])
-	setColor('ffffff')
+	setColor(color)
 }
 
 const selectTag = (
--- a/client/src/app/deck/deck-select.tsx
+++ b/client/src/app/deck/deck-select.tsx
@@ -61,29 +61,30 @@ function SelectDeck({
 
 	// STATE
 	const [savedDecks, setSavedDecks] = useState<Array<string>>(getSavedDecks)
-	function sortDecks(decks: string[]): Array<PlayerDeckT> {
-		return decks
-			.map((d: any) => {
-				const deck: PlayerDeckT = JSON.parse(d)
-				return deck
-			})
-			.sort((a, b) => {
-				if (settings.deckSortingMethod === 'Alphabetical') {
-					return a.name.localeCompare(b.name)
-				}
-				if (settings.deckSortingMethod === 'First Tag') {
-					const aHasTags = a.tags && a.tags.length > 0
-					const bHasTags = b.tags && b.tags.length > 0
-					if (!aHasTags && !bHasTags) return a.name.localeCompare(b.name)
-					if (!aHasTags && bHasTags) return 1
-					if (aHasTags && !bHasTags) return 0
-					const aFirstTag = a.tags![0]
-					const bFirstTag = b.tags![0]
-					return aFirstTag.localeCompare(bFirstTag)
-				}
-				//Default case so something is always returned
-				return 0
-			})
+	function parseDecks(decks: string[]): Array<PlayerDeckT> {
+		return decks.map((d: any) => {
+			const deck: PlayerDeckT = JSON.parse(d)
+			return deck
+		})
+	}
+	function sortDecks(decks: PlayerDeckT[]): Array<PlayerDeckT> {
+		return decks.sort((a, b) => {
+			if (settings.deckSortingMethod === 'Alphabetical') {
+				return a.name.localeCompare(b.name)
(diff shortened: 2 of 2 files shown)


alicetouhou:
Also, not in the scope of this PR, but we should eventually work on visual improvements for the tags. its not great. possibly @DevChimera can help out here

Sense101:
Huh good thing I didn't decide to work on this in the end.
I will review this when I get a chance.

And I'd be happy to work on visual improvements, I noticed a few things already. What specific things do you want improved?

alicetouhou:
> And I'd be happy to work on visual improvements, I noticed a few things already. What specific things do you want improved?

I just think my design sucks. Want to see what other people can come up with and improve from there

alicetouhou:
Let's make this an issue though if we're gonna discuss

client/src/app/deck/deck-select.tsx
```diff
@@ -194,8 +208,15 @@ function SelectDeck({
...
-		deckExists && setShowOverwriteModal(true)
-		!deckExists && saveDeckInternal(deck)
+		if (deckExists) {
+			setShowOverwriteModal(true)
+			return
+		}
+
+		saveDeckInternal(deck)
+		setSavedDecks(getSavedDecks())
+		setSortedDecks(sortDecks([...savedDecks, JSON.stringify(deck)]))
```
Lunarmagpie, line 218: Why are you stringifying then parsing later?

alicetouhou: The function takes in only strings... YEah this is really bad. I should refactor this part of the code so multiple behavior isn't handled in one function

Lunarmagpie: Please fix this... this function should not take a string.

alicetouhou: Resolved

Lunarmagpie:
LGTM

Lunarmagpie (approved):

alicetouhou:
That one change actually helped clean up a lot of the code. The only issue I have now is I need to set multiple states the same thing and I have no idea why.

Lunarmagpie:
You should fix that

client/src/app/deck/deck-select.tsx
```diff
@@ -395,7 +442,9 @@ function SelectDeck({
 			<AlertModal
 				setOpen={showDuplicateDeckModal}
 				onClose={() => setShowDuplicateDeckModal(!showDuplicateDeckModal)}
-				action={() => duplicateDeck(loadedDeck)}
+				action={() => {
+					duplicateDeck(loadedDeck)
+				}}
```
Sense101, line 447: Do we need these extra brackets here?

client/src/app/deck/deck-select.tsx
```diff
@@ -61,12 +61,14 @@ function SelectDeck({
 
 	// STATE
 	const [savedDecks, setSavedDecks] = useState<Array<string>>(getSavedDecks)
-	const sortedDecks = savedDecks
-		.map((d: any) => {
+	function parseDecks(decks: string[]): Array<PlayerDeckT> {
```
Sense101, line 64: This seems extremely overcomplicated to me. We're saving the decks as strings and continuously reparsing them. What should really be happening is we should have a tracked state for a PlayerDeckT array that gets updated using the useState stuff, and every time a change is made we should stringify and save decks to local storage. We should not need to parse from strings other than when we first load.

This was probably always set up wrong tbh. This whole thing could be simplified a lot, but right now it works so therefore I think we should just merge it, and then redo the whole file at a later date.

Sense101 (commented):
Reviewed. It's very spaghetti but the file was already such long spaghetti I don't think it matters. See review comments for more detail.

client/src/app/deck/deck-edit.tsx
```diff
@@ -128,7 +128,7 @@ const addTag = (
 	if (tags.length >= 3) return
 	if (tag.name.length === 0) return
 	setTags([...tags, tag])
-	setColor('ffffff')
+	setColor(color)
```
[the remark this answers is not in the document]
alicetouhou: I don't think this is in scope as I didn't modify the actual tag system with this PR, just a bug fix. But we can redo it in the future when I have time

PR #966, opened 2024-10-23
Make it possible for status effect descriptions to be dynamic

Lunarmagpie:
It would be best to implement this for one status effect to prove the system is good before going ahead and merging 
This PR additionally sets up `InstancedValue` so it can be used with status effects.


alicetouhou:
Ollie is the best candidate, but if course AoTCG isn't done yet. Maybe do skizz and show the total damage gained?

Lunarmagpie:
I'm thinking the multi turn and single turn attack disabling effects could be merged together possibly.

alicetouhou:
> I'm thinking the multi turn and single turn attack disabling effects could be merged together possibly.

How so? If it works go for it

Lunarmagpie:
> > I'm thinking the multi turn and single turn attack disabling effects could be merged together possibly.
> 
> How so? If it works go for it

I think it actually won't work.

alicetouhou:
Can we close this?

Lunarmagpie:
> Can we close this?

No, I want to compare the methods than close this after we merge

PR #967, opened 2024-10-23
Allow status effect descriptions to display differently on front end

alicetouhou:
also removes some duplicate code


Lunarmagpie (approved):
Looks reasonable

PR #969, opened 2024-10-27
Combine Sword and Armor Implementations

Lunarmagpie:
Combines sword and armor implementations to reduce the chance of bugs in the future!


alicetouhou:
awesome！this is a long time coming. LGTM

Screaper91 (approved):
Makes sense to me, I doubt the armor will be changed to have unique properties (original Gold breaking or S2 Diamond/Netherite) and if so it can be split back

Lunarmagpie:
> Makes sense to me, I doubt the armor will be changed to have unique properties (original Gold breaking or S2 Diamond/Netherite) and if so it can be split back

Almost certainly not! I think its a lot more likely armor will need to be changed because of a bug with some new interactions lol.

PR #971, opened 2024-10-30
Move all modals over to new modal system

Sense101:
The modal code has been expanded and improved to handle all possibilities we need it for, all modals have been converted over to the new system, and various small changes have been made to many modals to improve them.

Ready for review.


alicetouhou:
image
Selection becomes cut off in select cards modal

alicetouhou:
image
In the situation where there's one card, it's supposed to be centered. In addition, This card also expands on hover even though the selection size is 0. It's not actually selectable, though.

alicetouhou:
It seems you need to press escape twice to close any closable modal. I don't know why but the first escape press does not do anything

client/src/app/game/modals/change-hermit-modal.tsx
```diff
@@ -54,32 +54,29 @@ function ChangeHermitModal({closeModal, info}: Props) {
...
-						</>
-					)}
-				</div>
-			</div>
+		<Modal setOpen title="Change active hermit" onClose={handleNo}>
+			<Modal.Description>
+				{lastAction && (
+					<div className={css.turnEndNotification}>
+						<span className={css.infoIcon}>!</span>
+						{/* Hermit change is the last action of your turn. */}
```
Lunarmagpie, line 62: This comment doesnt add any useful information, meaning it will probably just become wrong or something

Lunarmagpie (approved):
Looks good overall! Need tests tho just to make sure the visuals don't have issues in the future. The card select modal is the most important one to have tests for.

Lunarmagpie (changes requested):
Meant to request changes not approve lol

client/src/app/game/modals/change-hermit-modal.tsx
Sense101: Not written by me, but can remove

alicetouhou:
I think there's a div with the id "modal" you can delete now. Found it while experimenting with tooltips

client/src/app/game/modals/change-hermit-modal.tsx
Sense101: Turns out it was somehow added by me, but fixed now.

Sense101:
Known issues: select cards modal needs minor fixes. Should be easy to do.

Lunarmagpie:
> Known issues: select cards modal needs minor fixes. Should be easy to do.

I believe we should skip this issue for now and handle it later because this is currently an issue on master.

Lunarmagpie (approved):
LGTM

Lunarmagpie:
Testing the select cards modal and ensuring it looks correct is a huge can of worms, considering its already wrong on master i don't think this needs to be done immediately

PR #972, opened 2024-10-30
Fix tooltip z-order

alicetouhou:
Fixes bug where tooltips would always display above modals


Lunarmagpie:
This would make tooltips display below the card selection modal

Lunarmagpie:
Nevermind

PR #973, opened 2024-10-30
Change Potion of Weakness description

Rvtar:
It should be more clear now.


alicetouhou:
This doesn't add clarity at all, "current" can either mean "the Hermit that is active during the current turn" OR "the Hermit that is active while this effect is applied. Considering how wordy it is to specify the first behavior, I would just change the behavior to be what people seem to find intuitive currently and leave the description the same.

Rvtar:
Seems like the old one already worked fine now with more basis for judgement.

PR #975, opened 2024-10-31
Make sure database is not disabled in default debug config

alicetouhou:
This made it so actions were not properly able to be executed on the online version


Lunarmagpie:
I think this config option shouldn't ever be read

alicetouhou:
Ok. I can set up the system in an alternate way so it does what this debug config option currently does *if* the database can't connect.

Lunarmagpie (approved):
Kansei de perfecto!

PR #977, opened 2024-11-01
Validate starter decks

Lunarmagpie:


alicetouhou:
I'm going to be annoying, if you're fixing this can you also make it so the decks just have the cards listed out instead of using code

alicetouhou:
Also if you are working on this, you could expand the system to return a name and icon we can modify, would be great

Lunarmagpie:
You're building a shed but I'll change it

alicetouhou:
Looks good

PR #978, opened 2024-11-01
Remove floating window from tooltips

alicetouhou:


Lunarmagpie (approved):
LGTM, has nice performance

Lunarmagpie:
Getting some strange behavior that needs to be investigated

PR #979, opened 2024-11-03
Remove Radix from Toasts

alicetouhou:
The behavior is almost identical, however:
- I reduced the time toasts show dramatically. I think it was way too long before. Clicking on them without dragging refreshes the time, if you want to read them longer. I think that makes sense.
- You can now double click to close a toast
- Toasts can now stack when multiple are showing at once


Lunarmagpie:
Looks like it works, but since this is being redone I think you should implement the queue for toasts or make them stack vertically (tbh i like this more). Additionally make a component test to make sure our toasts are styled correctly while you're at it.

alicetouhou:
I was actually able to fix the issues with timing and added double clicking

alicetouhou:
Going to see if i can get stacking working, but I might not be able to figure it out

alicetouhou:
This is done now

PR #982, opened 2024-11-04
Add drag modal for brush

alicetouhou:
Draft. Visuals aren't done and tests obviously fail


Screaper91 (commented):
The drag modal is hard to use when I tested it, the cards started offscreen and jumped around when I tried rearranging the two cards. The order of cards moved to the bottom also is determined by the order from the modal result while the description states "original order".

alicetouhou:
Second issue easy fix, but for the first do you mind sending me a video? It will make diagnosing the issue a lot easier.

client/src/app/game/modals/drag-cards-modal.tsx
```diff
@@ -0,0 +1,425 @@
...
+					</div>
+					<div className={css.subContainer}>
+						{[...startingLeftCards, ...startingRightCards].map((card, i) => {
+							return (
+								<DraggableCard
+									entity={card.entity}
+									draggedCard={draggedCard}
+									setDraggedCard={setDraggedCard}
+									cardInfo={cardsInfo}
+									key={i}
```
Screaper91, line 390: Just so this change doesn't get lost from our debugging
```suggestion
									key={card.entity}
```

PR #984, opened 2024-11-04
Prerender Cards with Playwright

Lunarmagpie:
This PR renders cards into webps using Playwright and Imagemagick. A github action is added so users do not need to run the script to do this on their system.


Lunarmagpie:
Besides for renaming some folders for clarity and docs, this should be done.

PR #986, opened 2024-11-05
Fix some issues with database

alicetouhou:
- Fix active deck not being properly set when a new user is created
- Fix user data resetting not working properly


Lunarmagpie (approved):
NGL I didnt run this but its probably good

PR #999, opened 2024-11-09
Deck issues

alicetouhou:
- Removes the overwrite modal, because it isn't necessary anymore
- Fixes the issue with the tag filter being able to be active when you create a new account
- Fixes decks being imported without their tags
- Fixes decks from local storage sometimes not being imported properly after your first time making an account
- Fixes Active decks not being set properly in all scenarios


Lunarmagpie (approved):
I have verified this to work

PR #1000, opened 2024-11-09
Fix UUID Issue

alicetouhou:
Fixes server crash being possible when a game with less than 2 players ends


Lunarmagpie (approved):
LGTM, I think this is good since Evil X games should not be saved

PR #1004, opened 2024-11-10
Improvements to mobile cards

alicetouhou:
- Fixes issue where cards would show the image URL on hold
- Fixes issue where cards in the deck builder would be randomly removed when you tried to scroll
- Fixes issue where it was confusing how to remove cards from the deck builder

### DIFF
--- a/client/src/components/card/mobile-card-component.tsx
+++ b/client/src/components/card/mobile-card-component.tsx
@@ -12,13 +12,13 @@ interface CardReactProps
 	> {
 	cards: Array<LocalCardInstance>
 	tooltipAboveModal?: boolean
-	onClick?: () => void
+	onSubtractionClick?: () => void
 	onAdditionClick?: () => void
 	small: boolean
 }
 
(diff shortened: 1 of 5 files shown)


client/src/components/card/mobile-card-component.tsx
```diff
@@ -12,81 +12,84 @@ interface CardReactProps
...
+						<img
+							className={css.headInList}
+							src={`/images/effects/${cards[0].props.id}.png`}
+						/>
+					)}
+					{cards[0].props.category === 'item' && (
+						<div>
+							<img
+								className={css.headInList}
+								src={`/images/types/type-${cards[0].props.id.split('_')[1]}.png`}
```
Lunarmagpie, line 59: Theres a function called `getTypeImage` or something I think you can use here. We use this image in a lot of places so it needs to be consistent.

Lunarmagpie (changes requested):

Lunarmagpie (approved):
LGTM. Tested it and it works well.

PR #1005, opened 2024-11-10
Fixes to active deck system

alicetouhou:
This PR should make it impossible to join a game if your active deck is not valid. Right now, there seems to still be some situations you can join with an invalid deck, and it uses a starter deck instead.

### DIFF
--- a/server/src/routines/matchmaking.ts
+++ b/server/src/routines/matchmaking.ts
@@ -47,11 +47,11 @@ function setupGame(
 	let game = new GameModel(
 		{
 			model: player1,
-			deck: player1.deck.cards.map((card) => card.props.numericId),
+			deck: player1.deck!.cards.map((card) => card.props.numericId),
 		},
 		{
 			model: player2,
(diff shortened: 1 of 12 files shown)


server/src/routines/matchmaking.ts
```diff
@@ -47,11 +47,11 @@ function setupGame(
 	let game = new GameModel(
 		{
 			model: player1,
-			deck: player1.deck.cards.map((card) => card.props.numericId),
+			deck: player1.deck!.cards.map((card) => card.props.numericId),
 		},
 		{
 			model: player2,
-			deck: player2.deck.cards.map((card) => card.props.numericId),
+			deck: player2.deck!.cards.map((card) => card.props.numericId),
```
Lunarmagpie, line 54: I dont think using the bang operator here is a very good idea

Lunarmagpie (dismissed):

server/src/routines/matchmaking.ts
alicetouhou: Fixed this

PR #1006, opened 2024-11-11
Fix importing local decks with Flint & Steel

Screaper91:


Lunarmagpie (approved):
that was an embarrassing mistake on my part 😭

PR #1007, opened 2024-11-11
Implement candle

Screaper91:


alicetouhou:
LGTM :)
However I would slightly modify the description to be more accurate: `Search your deck for any pair of cards. For one pair you find, draw one copy, discard the other, then shuffle.`
Does this look good to you?

Lunarmagpie (approved):
I really like this idea

Lunarmagpie (approved):

PR #1008, opened 2024-11-11
Implement TNT Minecart

Screaper91:


alicetouhou:
I think we should change the positive damage to 100hp, this would be pretty worth it for 2 tokens as an alternate to TNT.

Lunarmagpie:
Looks great!

Lunarmagpie (approved):

PR #1013, opened 2024-11-11
Fixes tags only being added to the first deck after being imported from local storage

alicetouhou:


Lunarmagpie (approved):
i dont know whats happening here but it looks correct

alicetouhou:
The mistake before was I changed the ID of the tag IN the array, so any future tags would not be found since their key was different, now it works properly

PR #1017, opened 2024-11-11
Add games count and detail endpoint

Tyrannicodin:


server/src/api/games.ts
```diff
@@ -24,6 +26,23 @@ function cancelGame(game: {
 	if (game.gameCode) delete root.privateQueue[game.gameCode]
 }
 
+function getPlayers(game: GameModel) {
+	return game.components.filter(ViewerComponent).flatMap((viewer) => {
+		if (viewer.spectator) return []
+		let player = viewer.playerOnLeft
+		return [
+			{
+				playerId: viewer.playerId,
```
Lunarmagpie, line 35: I don't think this is necessary to expose through the endpoint. Also, I think it would be better to return a dictionary with the keys `playerOne` and `playerTwo`.

Lunarmagpie (changes requested):

Lunarmagpie:
Looks good besides my one request.

Lunarmagpie (approved):

PR #1027, opened 2024-11-14
Fix deck validation issues

alicetouhou:


Lunarmagpie:
Im pretty sure validateDeck already checks this variable. If not you should be doing this check there instead.

alicetouhou:
Oh I guess I never broke this then. I'll close this

PR #1032, opened 2024-11-16
Add started spectating message

Lunarmagpie:
Adds message when players start spectating and when they stop spectating.


Sense101:
This could potentially allow someone not playing to spam the chat, no?

Lunarmagpie:
> This could potentially allow someone not playing to spam the chat, no?

They can do that already. Don't share spectator codes with people you don't trust I guess? This isn't an issue to worry about at all.

PR #1034, opened 2024-11-17
Fix EX boss clearing Amnesia too often

Screaper91:
Evil X boss no longer clears Amnesia too often when used with a single use attack.
_**Note:** Amnesia only works on the Evil X boss every other turn_

### DIFF
--- a/common/cards/boss/hermits/evilxisuma_boss.ts
+++ b/common/cards/boss/hermits/evilxisuma_boss.ts
@@ -328,7 +328,7 @@ const EvilXisumaBoss: Hermit = {
 				.forEach((card) => destroyCard(game, card))
 			removeImmuneEffects(game, component.slot)
 		})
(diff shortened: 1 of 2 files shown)


common/cards/boss/hermits/evilxisuma_boss.ts
```diff
@@ -328,7 +328,7 @@ const EvilXisumaBoss: Hermit = {
 				.forEach((card) => destroyCard(game, card))
 			removeImmuneEffects(game, component.slot)
 		})
-		let lastAttackDisabledByAmnesia = false
+		let lastAttackDisabledByAmnesia = NaN
```
Lunarmagpie, line 331: I don't think using `NaN` is a very good idea. I find its really confusing so it lead to bugs.

Lunarmagpie (changes requested):

Lunarmagpie (approved):

PR #1035, opened 2024-11-17
Bundle

Tyrannicodin:
Send help, I can't figure out how to make this work nicely - it needs to work with damage single uses, non-damage, confirmation, non-confirming etc.


alicetouhou:
LGTM!

common/cards/advent-of-tcg/single-use/bundle.ts
```diff
@@ -0,0 +1,110 @@
...
+		query.card.slot(query.slot.hand),
+		query.card.player(player),
+	)
+
+const Bundle: SingleUse = {
+	...singleUse,
+	id: 'bundle',
+	numericId: 251,
+	name: 'Bundle',
+	expansion: 'advent_of_tcg',
```
Tyrannicodin, line 28:
```suggestion
	expansion: 'advent_of_tcg_ii',
```

common/cards/advent-of-tcg/single-use/full_bundle.ts
```diff
@@ -0,0 +1,148 @@
...
+import {applyCard, applySingleUse} from '../../../utils/board'
+import {singleUse} from '../../defaults'
+import {SingleUse} from '../../types'
+
+const FullBundle: SingleUse = {
+	...singleUse,
+	id: 'full_bundle',
+	numericId: -2,
+	name: 'Full Bundle',
+	expansion: 'advent_of_tcg',
```
Tyrannicodin, line 14:
```suggestion
	expansion: 'advent_of_tcg_ii',
```

Tyrannicodin (commented):
This need to be updated

client/src/app/game/modals/attack-modal/attack-modal.tsx
Screaper91, line 1: This needs work because the single use still appears after Grianch flips heads on Naughty, even though the card was used up and won't be applied again

PR #1036, opened 2024-11-17
Lower fetch depth in CI

Lunarmagpie:
This should speed up CI a bit.


Lunarmagpie:
It seems like this isn't actually required if you aren't using `git push`.

PR #1040, opened 2024-11-21
Show rarities on effect card tooltips

alicetouhou:
additionally fixes rarities of advent cards to be more logical


Lunarmagpie (approved):

alicetouhou:
Should we show the rarities of item cards too?

Lunarmagpie:
> Should we show the rarities of item cards too?

It's up to you.

PR #1041, opened 2024-11-21
Add windows card renderer

Tyrannicodin:


Lunarmagpie:
Thank you! Can you switch this to use Imagemagick instead though?

Lunarmagpie:
One last thing, can you add a command `npm run client:render-cards-nt` to run this?

PR #1042, opened 2024-11-22
Changes to make API easier to use

Lunarmagpie:
Unfortunately breaking

### DIFF
--- a/API.md
+++ b/API.md
@@ -1,27 +1,53 @@
 # Hermitcraft TCG API Documentation
 
-- GET `/api/cards`
+- `GET /api/cards`
+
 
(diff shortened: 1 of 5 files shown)


API.md
```diff
@@ -1,27 +1,54 @@
...
+
 
 Returns a list of all card types and resources for them.
 
-- GET `/api/ranks`
+---
+
+- `GET /api/ranks`
+
+---
```
Tyrannicodin, line 19: I think this should this be after the /api/ranks description

Tyrannicodin, line 19: I think this should this be after the /api/ranks description

PR #1047, opened 2024-11-24
Remove radix from dropdown menu and improve it's behavior

alicetouhou:
- Removes radix from the project entirely (yay!)
- Adds more deck icons to choose from
- Adds check boxes to the expansion filter


Lunarmagpie:
Millions must use hand-rolled HTML and JS

PR #1058, opened 2024-11-26
Use Set Seed For Games

Lunarmagpie:
This is to help make it possible to run games on the client in the future. This is also helpful because this change makes tests 100% consistent as they are no longer reliant on RNG.


Lunarmagpie:
> please make the database actually use the seed

Done

alicetouhou:
LGTM

alicetouhou:
Actually, how long is the seed? There may be an issue if it's over 15 characters

Lunarmagpie:
> Actually, how long is the seed? There may be an issue if it's over 15 characters

Idk but I can make it 8

Lunarmagpie:
Probably good to merge now

PR #1059, opened 2024-11-26
Update API stats to be more in depth

alicetouhou:


Lunarmagpie:
Make sure to update the docs

PR #1062, opened 2024-11-27
Authenticate /api/stats/decks

Tyrannicodin:


Lunarmagpie:
There is one small thing I am worried about, how is the privileged endpoint planned to be used? If anyone gets access to the consumer of the endpoint I don't think it makes sense to allow this.

alicetouhou:
That is true, looking at decks below <20 wins is a bit useless anyway

PR #1065, opened 2024-11-27
Add a debug option to render cards dynamically instead of using images

alicetouhou:
This will make it easier to test visual changes to the cards

### DIFF
--- a/common/config/debug-config.example.js
+++ b/common/config/debug-config.example.js
@@ -20,4 +20,5 @@ export default {
 	autoEndTurn: false,
 	logAttackHistory: false,
 	disableRewardCards: false,
-}
+	showCardsAsHTML: false,
+};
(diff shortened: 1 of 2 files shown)


common/config/debug-config.example.js
```diff
@@ -20,4 +20,5 @@ export default {
 	autoEndTurn: false,
 	logAttackHistory: false,
 	disableRewardCards: false,
-}
+	showCardsAsHTML: false,
```
Lunarmagpie, line 23: I don't think this name makes a lot of sense

Lunarmagpie (changes requested):

common/config/debug-config.example.js
alicetouhou: What would you make it?

Tyrannicodin: How about `dynamicCardRenders`?

alicetouhou: i like `renderCardsDynamically`

Lunarmagpie: That sounds good

Lunarmagpie (approved):

PR #1066, opened 2024-11-27
Merge Upstream With AOTCG

Lunarmagpie:
This is actually going to make me cry


Lunarmagpie:
This should be the new aotcg branch. I will skip using github for the merge.

PR #1067, opened 2024-11-28
Improvements to `type-distribution` and `cards` endpoint, and refactor queries to be faster

alicetouhou:
Docs explain the changes made

### DIFF
--- a/API.md
+++ b/API.md
@@ -127,10 +127,16 @@ Returns the 20 top scoring decks that match the specified parameters. Each deck 
 
 -   `POST api/stats/type-distribution`
 
-Returns information for each type.
+Returns information for types. To make the analysis more accurate, this endpoint does not include decks with a winrate below 0.05% or above 0.95% in the data in analyzes.
 
--   `type` The name of the type.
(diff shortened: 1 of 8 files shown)


API.md
```diff
@@ -127,10 +127,16 @@ Returns the 20 top scoring decks that match the specified parameters. Each deck
...
 -   `POST api/stats/type-distribution`
 
-Returns information for each type.
+Returns information for types. To make the analysis more accurate, this endpoint does not include decks with a winrate below 0.05% or above 0.95% in the data in analyzes.
 
--   `type` The name of the type.
--   `usage` The percentage of decks that were played in games that include an item card of this type.
+-   `dualTypeFrequency` The percent of decks that are dual type.
+-   `monoTypeWinrate` The percent of games where a player used a mono type deck that the deck won.
+-   `dualTypeWinrate` The percent of games where a player used a dual type deck that the deck won.
```
Lunarmagpie, line 134: Would it make more sense to do `multiTypeWinrate` instead of `dualTypeWinrate`?

alicetouhou: It probably would make more sense

PR #1077, opened 2024-11-29
Support Showing Satus Effects in Card List/Select Cards Modal

### LINKED ISSUE #1012: Show card statuses in CardList for Smithing Table
#980 Added Smithing Table and allowed providing status effects to `<CardList/>`, it does not display on the client until the Attach is played but should be visible when in hand or in a select cards request (e.g. Spyglass)

Lunarmagpie:
This needs to be tested with Furnace on the AOTCG Branch

Closes #1012


Screaper91 (approved):
LGTM to be ready for next card drop

PR #1078, opened 2024-11-29
Various front end fixes

alicetouhou:
- Fix decks not using name you enter when being imported
- Fix imported decks dropdown being too long
- Fix imported decks warning message not resetting after re-opening the window
- Fix imported decks import button being activated, even when the deck wasn't valid, for deck codes with a length of 7.
- Fix tag filter dropdown opening downwards instead of upwards
- Fix icons in dropdown buttons not being centered
- Improve design of checkboxes, along with fixing centering on some devices.
- Fix the Emoji button in chat window not having Bdub's head centered


Lunarmagpie:
Just to confirm, this is finished?

alicetouhou:
Yes this is finished

PR #1082, opened 2024-11-30
Advent Of TCG 2

Lunarmagpie:
I actually would prefer if we didn't squash this. If theres any issues being able to look though the history would be good.


card-prerender/render.spec.tsx
Screaper91, line 1: These changes need to be removed, it hides the token cost for `${card.id}-with-tokens` and breaks the `architectfalse_common` images

Lunarmagpie: My bad, that rebase was a little terrible

PR #1102, opened 2024-12-03
Fix Typo in Pearl Description

Lunarmagpie:


alicetouhou:
aiyaa not a typo. the whole project uses "can not"

PR #1107, opened 2024-12-09
fix Aussie Ping flipping on consecutive turns after heads

Screaper91:
Fixes Aussie Ping not applying Immune status if either player brought a rare Skizzleman


alicetouhou:
WOW that is a crazy interaction. Congrats on finding it

Lunarmagpie:
You are THE GOAT

PR #1108, opened 2024-12-09
Add Priorities to Coin Flip Hook

Lunarmagpie:
This is the first step to removing `addBefore`. Screaper's system is so much better for maintainability so I want to move the entire project to it!


Lunarmagpie:
@Screaper91 could you check over this PR?

Screaper91 (approved):
This should interact the same way, though `onCoinFlip` should maybe change its signature from `(card: CardComponent, coinFlips: Array<CoinFlip>) => Array<CoinFlip>` since the returned value is always ignored

Lunarmagpie:
> This should interact the same way, though `onCoinFlip` should maybe change its signature from `(card: CardComponent, coinFlips: Array<CoinFlip>) => Array<CoinFlip>` since the returned value is always ignored

Nice catch!

PR #1115, opened 2024-12-17
Refactored redstone-torch.ts

linco95:
I was practicing some refactoring techniques. Thought I might open a PR to share the results if you feel like this would improve the code. 

I did this while the redstone torch was still a PR (#1019), but I've updated the file with the upstream changes since I create the initial [PR](https://github.com/alicetouhou/hc-tcg/pull/8)

### DIFF
--- a/common/cards/advent-of-tcg/single-use/redstone-torch.ts
+++ b/common/cards/advent-of-tcg/single-use/redstone-torch.ts
@@ -1,12 +1,19 @@
 import assert from 'assert'
-import {CardComponent, ObserverComponent} from '../../../components'
+import {
+	CardComponent,
+	type ObserverComponent,
+	type PlayerComponent,
+} from '../../../components'
 import query from '../../../components/query'
-import {GameModel} from '../../../models/game-model'
+import {type CardEntity} from '../../../entities'
+import {type GameModel} from '../../../models/game-model'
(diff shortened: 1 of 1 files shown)


Lunarmagpie:
Hello, I don't think we will be merging this because I dont think its a good idea to refactor without a big reason, but since you did put the work in ill leave some comments so you can have some feedback on your changes.

Lunarmagpie (commented):
A big issue I have with the code is that it is too long. Compared to the old version I have to jump around much more to find where things are happening. The two reasons for this are pulling out the onAttach function to be defined at the root of the file, which isn't really a deal breaker, but the larger issue is that you are making each step into its own function. For example `getTntCardEntitiest` requires me to jump around even though its only used in one place and the code inside of it is already pretty clear because of the variable names used and the description of the card. I have the same issues with the handle result function. I have to jump around the code even though I can already understand what the code does without help from the function names.
Another issue is having a function defined inside another function tells me you are probably capturing a variable. If you are not capturing a variable, simply define the function at the root level of the file,

alicetouhou:
I agree with this. It's much more confusing. Compared to the old version, I have a lot of a harder time figuring out what's going on. I don't really understand the motivation for changing what you did.

PR #1128, opened 2025-01-03
Change Armor to Creo Armor

zunda-arrow:
Changes armor to Creo implementations.
- Pick requests that are forced by hermits are skipped if the player can not be knocked back
- Egg can still be played when the opponent has netherite armor, so the pick request is not skipped.

This change is being done to prevent confusion about the online cards and to shake up the game a little bit.


alicetouhou:
Merging this but we may still make changes afterwards

PR #1130, opened 2025-01-04
Disable Iskall Cards

zunda-arrow:
Disable iskall cards. If balance changes are being made, I would like to do this.


alicetouhou:
Approved

PR #1132, opened 2025-01-05
Active decks rework

alicetouhou:
Instead of the active deck of a player being the full deck information, the active deck is now only saved/selected as the deck code. This means that it can be selected instantly and not when the deck info is retrieved from the database. When a player queues into matchmaking, the full deck info is retrieved from the database. This ensures players are never matched into games with the wrong deck.

**Bug Fixes**
- Fixes the deck you queue into a game with sometimes being outdated if you edited the deck on another device.
- Fixes the wrong deck being used when players with bad ping quickly queue into a game after changing their deck.


zunda-arrow:
Nice work! I think this is a better system than the previous one.

zunda-arrow:
We also should consider testability with this patch, and additionally running without the database connected. I think it would be better to use the entire deck and not the deck code.

zunda-arrow (approved):

PR #1153, opened 2025-01-15
Make Lightning Rod not Discard after Target Block Bypass

Rvtar:
Made Target Block Ignore Lightning Rod


zunda-arrow:
Please write tests for this. Otherwise looks good.

zunda-arrow:
It would probably be better to make lightning rod *not* discard if it detects that the target changes after it sets it's target.

Rvtar:
That would definitely be a more general solution.

zunda-arrow:
Hmm, we could probably count if "redirect" shows up in the attack history more than once but that might be too complicated.

zunda-arrow:
I think we can do something like this. It would not work with grianch though I think.
```ts
observer.subscribeWithPriority(
	game.hooks.afterAttack,
	afterAttack.UPDATE_POST_ATTACK_STATE,
	(attack) => {
		if (
			attack.getHistory().filter((x) => x.type === 'redirect').length > 1
		) {
			used = false
		}
	},
)
```

zunda-arrow:
I thought about this more.
The current implementation is good. This PR just needs tests.

Rvtar:
Hope I did it right.

zunda-arrow (approved):
LGTM

PR #1160, opened 2025-01-19
Re-texture Iskall85 to Farmer Beef

Tyrannicodin:


Sense101 (approved):

alicetouhou (changes requested):
I think we need to switch farmer beef over to an alter egos card. Otherwise it won't be consistent with the rest of the cards in the game.

alicetouhou (approved):

PR #1165, opened 2025-01-21
Fix disabling again

alicetouhou:
I made a mistake before so cards were not greyed out when people's cached images updated, this should fix it


zunda-arrow:
This works, but im not really sure that I like the cursor changing to the disabled one

PR #1168, opened 2025-01-21
Negative attack to heal

Tyrannicodin:
I have no idea if this is good, it feels like there should be a better way than just removing a load of checks.


common/cards/advent-of-tcg/single-use/inverter.ts
```diff
@@ -0,0 +1,33 @@
...
+import query from '../../../components/query'
+import {beforeAttack} from '../../../types/priorities'
+import {singleUse} from '../../defaults'
+import {SingleUse} from '../../types'
+
+const Inverter: SingleUse = {
+	...singleUse,
+	id: 'inverter',
+	name: 'Inverter',
+	expansion: 'default',
```
zunda-arrow, line 11: This should probably be in the debug folder or similar

zunda-arrow (changes requested):

common/cards/advent-of-tcg/single-use/inverter.ts
Sense101: agreed

common/components/row-component.ts
```diff
@@ -95,8 +95,7 @@ export class RowComponent {
 	public damage(amount: number) {
 		// Deduct and clamp health
 		if (this.health === null) return
-		const newHealth = Math.max(this.health - amount, 0)
-		this.health = Math.min(newHealth, this.health)
+		this.health = this.health - amount
```
Sense101, line 98: This now needs to be clamped to not heal more than the hermit's max health.

common/components/row-component.ts
```diff
@@ -95,8 +95,7 @@ export class RowComponent {
...
 	}
 
 	public heal(amount: number) {
```
Sense101, line 101: This method is now unneeded, and all hermits that use it should be moved over to negative damage

Sense101 (changes requested):
Overall I think it is mostly as simple as what you've done, other than the changes I mentioned. Relevant comments should probably be updated to reflect that negative damage is an option, and I think there may be some situations where exactly 0 damage should still be checked for, but I couldn't find anywhere here.

The only other thing then is going through all cards that heal and updating them.

common/cards/advent-of-tcg/single-use/redstone-torch.ts
```diff
@@ -82,7 +82,7 @@ const RedstoneTorch: SingleUse = {
 							target: opponentPlayer.activeRowEntity,
 							type: 'effect',
 							log: (values) =>
-								`${values.player} detonated their ${tntCardEntities.length} $e${TNT.name}$ with $e${RedstoneTorch.name}$ to attack ${values.target} for ${values.damage} damage `,
+								`${values.player} detonated their ${tntCardEntities.length} $e${TNT.name}$ with $e${RedstoneTorch.name}$ to attack ${values.target} for ${values.health} ${values.healing ? 'healing' : 'damage'} `,
```
zunda-arrow, line 85: I don't think this check should be necessary since TNT should never heal. I think the same for the rest of these checks.

zunda-arrow (changes requested):

Sense101:
> I am hesitant about running negative attacks through attack hooks, where it is harder to guarantee they are run before Wolf/Thorns and may trigger listeners extra times depending on how the `AttackModel` is initialized.

I kind of agree with this, would it be so bad to make a separate set of hooks and model for healing? I find myself a bit confused by negative damage tbh, and it could be confusing to new contributors.

zunda-arrow:
I feel we should close this PR.
Combining the concept of attacking and healing feels to me it creates a lot of problems and doesn't help fix much. Thanks for the work on this though! This PR is still great research in the way to improve the healing system that we wouldn't have been able to figure out without this work.

PR #1170, opened 2025-01-23
Fix clamping row health to hermit's max after using Ladder

Screaper91:
- Fixes healing a Hermit above max health reducing health to their max health
- Adds test to prevent re-introducing bug where attacks could "deal extra damage" if `row.health - attack.calculateDamage()` was greater than `hermit.props.health`


zunda-arrow (approved):
The code looks good, im just not sure if ladder is supposed to function like this

alicetouhou:
This looks good to me

PR #1173, opened 2025-01-25
Allow Coin Head To Be Settable In Player Model

zunda-arrow:
I believe this does the bulk of the work for custom coin heads.
Additionally sets Evil X's coin to use his head.


Tyrannicodin (approved):
LGTM!

zunda-arrow:
I realized I will need to adjust this to allow the coin "border" to be settable as well.

zunda-arrow:
I think this is done.

PR #1176, opened 2025-01-27
Add `prizeCard` boolean to cards

zunda-arrow:
This adds a `prizeCard` boolean to cards to check if they are, well, a prize card

### DIFF
--- a/server/src/routines/game.ts
+++ b/server/src/routines/game.ts
@@ -340,7 +340,7 @@ function checkHermitHealth(game: GameModel) {
 
 				// prize card
 				if (game.settings.disableRewardCards) continue
-				let prizeCard = game.components
+				const prizeCard = game.components
 					.filter(
 						CardComponent,
 						query.card.slot(query.slot.deck),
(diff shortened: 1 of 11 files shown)


server/src/routines/game.ts
```diff
@@ -338,17 +338,20 @@ function checkHermitHealth(game: GameModel) {
...
 					.filter(
 						CardComponent,
 						query.card.slot(query.slot.deck),
 						query.card.player(playerState.entity),
 					)
 					.sort(CardComponent.compareOrder)
 					.at(0)
-					?.draw(playerState.opponentPlayer.entity)
+				if (prizeCard) {
+					prizeCard.draw(playerState.opponentPlayer.entity)
```
Tyrannicodin, line 352: I think this should be the other way round, so that when onChangeSlot is called it's marked as a prizeCard already.

Tyrannicodin (changes requested):

Tyrannicodin (approved):

PR #1188, opened 2025-01-31
Move Achievements to `GameController` Class

zunda-arrow:
This allows achievements to only be run on the server, even when I have the game run on the client at the same time in the future.


alicetouhou:
LGTM

Tyrannicodin (approved):

PR #1189, opened 2025-01-31
Implement Basic Fuzz Testing

zunda-arrow:
This is a draft for fuzzy testing games. The idea is to play two AIs against each other in a bunch of game and verify that the game never gets into a valid state. For example, there should NEVER be a pick request with no options.
The second goal is to verify that asserts never trigger. Asserts are great, but only when you have them along with fuzzing.

This should be a huge help to the game's stability once it is implemented.

**Bug Fixes**
Looting now asks for you to pick a card before attacking. This is to fix looting creating a pick request with no choices after the user knocks out the opponent's active hermit.


server/src/routines/game.ts
```diff
@@ -557,11 +555,9 @@ function* turnActionsSaga(con: GameController, turnActionChannel: any) {
 		con.game.battleLog.sendLogs()
 
 		const playerAI = getPlayerAI(con.game)
-		if (playerAI && !playerAISagaRunning) {
+		if (playerAI) {
```
zunda-arrow, line 558: I can't figure out what the `!playerAISagaRunning` code was even there for so I removed it.

zunda-arrow: I think this was to prevent multiple turn action sagas from starting when evilx yielded multiple actions, but I'm not sure why that would be necessary with the current system

PR #1190, opened 2025-01-31
Re-write updates to use github releases and get locally

Tyrannicodin:


client/src/components/updates/updates.tsx
```diff
@@ -1,57 +1,70 @@
...
+}
+
 export function UpdatesModal({onClose}: UpdatesModalProps) {
-	const updates = Object.values(useSelector(getUpdates))
-	console.log(updates)
-	const latestUpdateElement = useRef<HTMLLIElement>(null)
+	const [data, setData] = useState<Update[]>()
+
+	async function getReleases() {
+		const res = await fetch('https://api.github.com/repos/hc-tcg/hc-tcg/releases')
```
zunda-arrow, line 22: It would be better to only fetch this one time when the server starts up

PR #1206, opened 2025-02-06
Fix card type being marked as playable incorrectly 

zunda-arrow:
Fixes card action A being marked as playable when you have a card that is of two types, A, B, but could only be played in slot B.

### DIFF
--- a/server/src/routines/game.ts
+++ b/server/src/routines/game.ts
@@ -245,16 +245,48 @@ function getAvailableActions(
 
 				if (pickableSlots.length === 0) return reducer
 
-				if (card.isHealth() && !reducer.includes('PLAY_HERMIT_CARD')) {
+				if (
+					card.isHealth() &&
+					!reducer.includes('PLAY_HERMIT_CARD') &&
+					game.components.find(
+						SlotComponent,
+						card.props.attachCondition,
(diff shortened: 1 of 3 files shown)


server/src/routines/game.ts
```diff
@@ -245,16 +245,48 @@ function getAvailableActions(
...
 					reducer.push('PLAY_HERMIT_CARD')
 				}
-				if (card.isAttach() && !reducer.includes('PLAY_EFFECT_CARD')) {
+				if (
+					card.isAttach() &&
+					!reducer.includes('PLAY_EFFECT_CARD') &&
+					game.components.find(
+						SlotComponent,
+						card.props.attachCondition,
+						query.slot.attach,
```
Screaper91, line 265: The unit test fails because String is an attach card that can be placed in an item slot (even when the opponent has no available attach slots)
```suggestion
						query.some(query.slot.attach, query.slot.item),
```

zunda-arrow: Thanks for finding the bug! I don't like the solution so I am going to try something else.

zunda-arrow: The solution in the PR right now passes all unit tests and fixes the fuzz testing errors.

PR #1220, opened 2025-02-09
Fix checkbox on chromium browsers

alicetouhou:
just made the image smaller and it seems to work


zunda-arrow (commented):
I don't have chrome installed so I trust you

zunda-arrow (approved):

PR #1229, opened 2025-02-10
New game end screen

Tyrannicodin:
- [ ] Connect view replay button
- [x] Disable cards, slots etc. after game
- [ ] Implement rematch


Sense101:
this should be going to ux-rework, I think

Tyrannicodin:
Overriden by #1255

PR #1230, opened 2025-02-10
Add copied code toasts

Tyrannicodin:


zunda-arrow:
Nice idea!

alicetouhou:
LGTM! Although, the image seems a bit random. Maybe we can use the book image for it? Seems maybe more code related

Sense101:
This works fine, and if we go with it I would also recommend changing the image to just the "copy icon" or something

In my head I was picturing something more like the discord copy popup:
image

Sense101:
Also this should be merge to ux-rework

PR #1234, opened 2025-02-12
Re-name "Keep name and icon private" to "Make name and icon public"

Tyrannicodin:
This makes it more consistent with how the database works, where it sets show_info to False by default.


zunda-arrow (approved):

alicetouhou:
Thank you for finally fixing this

Sense101 (approved):

PR #1235, opened 2025-02-13
Rework session-saga.ts, and database functions

alicetouhou:
This PR updates session saga to allow for players to be automatically logged in.
Players are now added using the API, and sockets always have a player UUID.
In addition, I'm reworking a lot of the database functions to be neater and work better with the new way players are added to the database.
Also adds icons to starter decks.
I'm not working on the front-end login screen improvements in this PR


Sense101:
Note: this also adds icons to starter decks
We ideally want all changes noted down so we can have a massive changelog, and this one would likely have flown under the radar

alicetouhou:
Added it
Next time, you can just edit the description instead of leaving a comment

client/src/logic/game/game-saga.ts
```diff
@@ -253,7 +252,6 @@ function* gameSaga(initialGameState?: LocalGameState) {
 				type: localMessages.GAME_END_OVERLAY_SHOW,
 				outcome,
 			})
-			yield call(updateAchievements, socket)
```
Sense101, line 256: Why is this no longer needed?

client/src/logic/session/session-saga.ts
Sense101, line 1: This file has too many changes to easily review, but I'm sure it's good, and we can always make fixes after

client/src/socket.ts
```diff
@@ -16,5 +17,8 @@ export function newSocket() {
 		console.log('[message]', event, payload)
 	})
 
+	console.log(socket)
```
Sense101, line 20: This needs to be removed

server/src/api/index.ts
```diff
@@ -26,6 +27,21 @@ import {
...
+		res.statusCode = ret[0]
+		res.send(ret[1])
+	})
+
+	app.post('/api/createUser/', async (req, res) => {
+		const username = req.get('username')
+		let ret = await createUser(username)
+		res.statusCode = ret[0]
+		res.send(ret[1])
+	})
```
Sense101, line 43: Does this check to make sure a user with the same credentials doesn't already exist?

server/src/root-model.ts
```diff
@@ -73,6 +73,11 @@ export class RootModel {
 	}
 	public addPlayer(player: PlayerModel) {
 		this.players[player.id] = player
+		const time = Date.now()
+		const date = new Date(time)
+		console.log(
+			`${date.toLocaleTimeString('it-IT')}: Player [${player.uuid}] has logged in.`,
+		)
```
Sense101, line 80: I'm not sure if this is the best idea for console readability, although we're long past the times when we used to check it to see how many players were online

server/src/routines/player.ts
```diff
@@ -68,13 +69,22 @@ export function* playerConnectedSaga(
 				messages: game?.chat,
 			})
 		} else {
-			console.log('invalid player connected')
+			const time = Date.now()
+			const date = new Date(time)
+			console.info(
+				`${date.toLocaleTimeString('it-IT')}: Invalid player connected.`,
+			)
```
Sense101, line 76: Same potential issue as above

Sense101 (changes requested):
Looks good mostly. What changes does this make to how decks are stored?

server/src/api/index.ts
zunda-arrow: `createUser` can NEVER repeat credentials so this is not an issue

server/src/db/db.ts
```diff
@@ -174,29 +176,65 @@ export class Database {
...
-				[username, minecraftName, secret, this.bfDepth],
+				[username, username, secret, this.bfDepth],
 			)
+
+			const playerUuid: string = user.rows[0]['user_id']
+
+			const starterPack = getStarterPack()
+
+			const deckInfo = {
+				name: 'Starter Deck',
```
Tyrannicodin, line 193: Since we have names for started decks, this could be name - Starter Deck instead maybe.

server/src/routines/matchmaking.ts
```diff
@@ -280,6 +281,7 @@ function* gameManager(con: GameController) {
 				turnActionsBuffer,
 				con.gameCode,
 			)
+			yield* sendAfterGameInfo(gamePlayers)
```
Tyrannicodin, line 284: This should be called anyway, to make sure Evil X works with achievements for example.

server/src/db/db.ts
```diff
@@ -174,29 +176,65 @@ export class Database {
...
-		minecraftName: string | null,
-	): Promise<DatabaseResult<User>> {
+	public async insertUser(username: string): Promise<DatabaseResult<User>> {
 		try {
 			const secret = (await this.pool.query('SELECT * FROM uuid_generate_v4()'))
 				.rows[0]['uuid_generate_v4']
 			const user = await this.pool.query(
 				"INSERT INTO users (username, minecraft_name, secret) values ($1,$2,crypt($3, gen_salt('bf', $4))) RETURNING (user_id)",
-				[username, minecraftName, secret, this.bfDepth],
+				[username, username, secret, this.bfDepth],
```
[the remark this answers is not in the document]
zunda-arrow: I think the current code is more clear than this suggestion

server/src/api/index.ts
alicetouhou: Yeah. The chance of creating a UUID that already exists is 0 for all practical purposes

alicetouhou:
> Is there a specific reason for switching to endpoints for user authentication?

Yes. Now we don't create a socket until AFTER the user is logged in, so all sockets can have a username, uuid, and only ever exist for authenticated players. This makes a lot of the server code a lot neater

server/src/db/db.ts
alicetouhou: The reason I didn't do this is the starter deck names are for testing purposes only. It helps identify which decks are not valid. I feel like if you see a name like "PvP Deck #2" it would make a lot less sense than "Starter Deck".

server/src/db/db.ts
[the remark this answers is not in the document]
alicetouhou: eh, I think tyrannicodin is right, there's no reason for this function to take in the Minecraft name as an argument anymore

client/src/logic/session/session-saga.ts
zunda-arrow: I'm not sure what the purpose of this comment is?

client/src/logic/socket/socket-saga.ts
```diff
@@ -76,6 +78,8 @@ function* socketSaga(): SagaIterator {
 	const socket = yield* select(getSocket)
 	const session = yield* select(getSession)
 
+	if (!socket) throw new Error('The socket should be defined at this point.')
```
zunda-arrow, line 81: i think it would be a good idea to add when the socket was supposed to be defined

server/src/routines/player.ts
```diff
@@ -138,12 +149,9 @@ export function* updateMinecraftNameSaga(
...
 	const player = root.players[playerId]
 	if (!player) return
-	player.setMinecraftName(minecraftName)
+	player.minecraftName = minecraftName
 
-	broadcast([player], {
-		type: serverMessages.NEW_MINECRAFT_NAME,
-		name: player.minecraftName,
-	})
+	yield* setMinecraftName(player.uuid, minecraftName)
```
zunda-arrow, line 154: What would happen if the user was logged in on two devices and changed their minecraft name. I do not think this would work properly but im probably only a LOGAF 5 for fixing it.

zunda-arrow (changes requested):

client/src/logic/socket/socket-saga.ts
alicetouhou: Into the assertion message? Sure. 
I think I'll make it a constant too, so I can make sure it's always equal

server/src/routines/player.ts
alicetouhou: If you change your name on another device, it will work when you refresh or open a new tab

server/src/db/db.ts
Tyrannicodin: That's reasonable

server/src/routines/matchmaking.ts
Tyrannicodin: Actually, I think I got mixed up here, never mind

alicetouhou:
> Looks good mostly. What changes does this make to how decks are stored?

It doesn't make any changes

zunda-arrow (approved):
Just fix CI

server/src/routines/player.ts
zunda-arrow: Ok, thats reasonable. I just wanted to make it functioned how you expected.

PR #1237, opened 2025-02-14
Add Docker Compose File For Development

zunda-arrow:
This file needs to rebiuld the entire client when there is a change right now which is kinda slow, but I cant figure out how to not do that.


zunda-arrow:
This works but I want to get vite working so it's easier to work with

zunda-arrow:
Works perfectly now!

PR #1238, opened 2025-02-15
Make Lightning Rod No Longer Discard from 0-Damage Attacks or Missed Attacks

Rvtar:


alicetouhou:
I haven't checked this works yet, but I trust it will once the unit tests pass
Also, turtle shell should probably work the same way now as well

Rvtar:
But it only stays for one turn, what difference does it make?

PR #1239, opened 2025-02-15
Games over time stats

Tyrannicodin:
@alicetouhou could you use your database to more thoroughly test this?
Also, it probably needs some looks work, but I can do that once I verify the stats are working.


alicetouhou:
Sure I'll check it out tomorrow

alicetouhou:
Looks good, I would just work on the styling to make the graph a bit shorter so it fits without scrolling, and make the bars filled in

alicetouhou:
Also, i think you can hide the timestamp for hour when the interval is one day or greater. it's a bit confusing to look at

Tyrannicodin:
Just need to rebase, then should be ok

PR #1241, opened 2025-02-16
Add "Use Like a Hermit" Achievement

zunda-arrow:
"Survive with an armor stand as your active hermit for three rounds."
Cosmetic reward is the "Certified Zombie" title


zunda-arrow:
should be ready to merge

PR #1251, opened 2025-02-19
Up CPU To Shared 4x

zunda-arrow:
Increase our allowed CPU to 4x the current compute. This will give us wiggle room if we get many more players. This also increases our ram.


alicetouhou:
looks good, assuming you read the docs I'm sure it works

zunda-arrow:
> looks good, assuming you read the docs I'm sure it works

The docs have been read thoroughly

PR #1254, opened 2025-02-19
Final achievements

Tyrannicodin:
This pr closes https://github.com/hc-tcg/hc-tcg/issues/1231 for a total of 25 initial achievements.


common/achievements/sheep-starer.ts
```diff
@@ -0,0 +1,57 @@
...
+
+const zedaphCards = [ZedaphPlaysCommon, ZedaphPlaysRare]
+
+const SheepStarer: Achievement = {
+	...achievement,
+	numericId: 14,
+	id: 'sheep_starer',
+	name: 'Sheep Starer',
+	icon: '',
+	description: 'Keep a red HP zedaph alive for three consecutive turns.',
```
zunda-arrow, line 18:
```suggestion
	description: 'Keep a red HP (100 - 0) Zedaph alive for three consecutive turns.',
```
I think it would be a good idea to put the numbers in

zunda-arrow (approved):
I think this is good. I looked over it quickly. We can add tests in the future.

PR #1255, opened 2025-02-19
Game end screen clean

Tyrannicodin:
This overrides #1229


Sense101:
Would recommend you remove the change to the play select screen, as I'll be working on that and it will cause conflicts

PR #1261, opened 2025-02-21
Add Levels To Achievements and Type Win Achievements

zunda-arrow:
Add levels to achievements, to make it easier to program achievements with multiple stages or for us to add stages to an achievement in the future. Also adds type win achievements.


Tyrannicodin (approved):
Looks good!
I didn't actually check the database, but if your tests pass it's probably fine.

PR #1264, opened 2025-02-21
SUStainable Achievement

zunda-arrow:
Adds an achievement for discarding 100 cards with composter. My idea for the reward is "green" heart or "plant" heart but I would need help from someone like ChimeraDev for that, so i want approval of the idea first.


alicetouhou:
You should add the cosmetic in this PR as well. And I think just plain emerald colored hearts would be best

Screaper91 (approved):

PR #1274, opened 2025-02-23
Fix e2e Tests

zunda-arrow:
This PR fixes the component tests so they pass. Fortunately all the logic was correct so just what where the buttons were needed to be updated.


zunda-arrow:
They pass!

PR #1275, opened 2025-02-24
Add "Eye of the Spider" Achievement and Reward

zunda-arrow:


zunda-arrow:
I forgot to handle milk bucket and need to fix that.

Tyrannicodin (approved):

PR #1279, opened 2025-02-24
Custom Message For Opponent Disconnects

zunda-arrow:
I think this will also add disconnects as their own type in the database which should help us gather important data.


Tyrannicodin (approved):
Merge conflicts

PR #1281, opened 2025-02-24
Add riddle to screen

alicetouhou:
This PR adds the riddle to the main menu


zunda-arrow (approved):
Looks reasonable, after this update let's change it to be a "community interaction modal" that tells people we have a discord and the project is open source

alicetouhou:
Well, there's no room for it on the menu, so we'd need to change where it is located, but yeah

PR #1282, opened 2025-02-24
Update icon to be property of achievement level

Tyrannicodin:
Also, make achievements page return to correct page


zunda-arrow:
I disagree with this method.
I think we should get the icon by checking what an achievements reward is when we display it. I am happy to implement my idea right now so we can compare the options.

Tyrannicodin:
I think there was a reason we didn't do that, I'd have to look back at conversations with Benji to be sure.

PR #1284, opened 2025-02-24
Alternitive Method To Get Achivement Icons

zunda-arrow:
This is an alternative to #1282


Tyrannicodin (approved):
I have no preference between the two, but if we do merge we need to remove the icon attribute from achievements.

zunda-arrow:
Ah yes! I forgot.
I strongly prefer my method because we don't need to set up the cosmetic in two places.

PR #1295, opened 2025-02-25
Add Hurtin Hermits Achievement

zunda-arrow:


common/achievements/hurtin-hermits.ts
```diff
@@ -0,0 +1,15 @@
+import {achievement} from './defaults'
+import {Achievement} from './types'
+
+const HurtinHermits: Achievement = {
+	...achievement,
+	numericId: 41,
+	id: 'hertin-hermits',
+	levels: [{name: "Hurtin' Hermits", description: 'Win 1 game.', steps: 1}],
```
Tyrannicodin, line 8: I'm guessing this should be Lose 1 game?

Tyrannicodin (changes requested):

common/achievements/hurtin-hermits.ts
zunda-arrow: I agree

Tyrannicodin (approved):

PR #1298, opened 2025-02-25
Set username min length to 3 to fix issues

alicetouhou:


client/src/app/main-menu/cosmetics.tsx
```diff
@@ -334,6 +334,7 @@ function Cosmetics({setMenuSection, page}: Props) {
 								<Button
 									onClick={() => {
 										if (!usernameRef.current) return
+										if (usernameRef.current.value.length < 3) return
```
zunda-arrow, line 337: You probably need to have an error message

zunda-arrow (changes requested):

PR #1304, opened 2025-02-27
Remaches and game end screen

alicetouhou:
I'm pretty close, I just need to fix the rematch window and add your last achievements showing when a game is over


Tyrannicodin (approved):
I've tested this and it all seems to work, code looks good too.

PR #1312, opened 2025-03-01
Fix Many Achievement Bugs

zunda-arrow:
This fixes a but where achievements would sorta half save if you did not complete a level yet. It also makes sure the progress bar in the achievements component uses the right number.


server/src/db/db-reciever.ts
```diff
@@ -429,6 +429,7 @@ export function* getDeck(code: string) {
 
 export function* updateAchievements(player: PlayerModel) {
 	assert(root.db.connected, CONNECTION_ASSERTION_MSG)
+	console.log(player.achievementProgress)
```
zunda-arrow, line 432: Reminder to remove this

Tyrannicodin (approved):

PR #1317, opened 2025-03-02
Attept to fix issues with replays

alicetouhou:
This PR creates another test to ensure replays are working properly. I also made the background of the overview darker so we can have an easier time looking at it
In addition, i've temporarily disabled huffman compression to see if that was the issue.


zunda-arrow:
I think it would be a good idea to add a version number so you can add a new encoding version in the future.

alicetouhou:
There already is 1, the first byte is the version number

PR #1319, opened 2025-03-02
Make The Zero Coin Flip Bug Impossible

zunda-arrow:
I think this is a good change?

### DIFF
--- a/common/utils/coinFlips.ts
+++ b/common/utils/coinFlips.ts
@@ -36,7 +36,7 @@ export function flipCoin(
 	times: number = 1,
 	currentPlayer: PlayerComponent | null = null,
 ): Array<CoinFlipResult> {
-	if (times == 0) {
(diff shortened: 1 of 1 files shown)


alicetouhou:
I think it's good, but i would also modify frenchralis to not try to flip 0 coins, because the code is easier to read

zunda-arrow:
> I think it's good, but i would also modify frenchralis to not try to flip 0 coins, because the code is easier to read

I disagree. I think its better to flip 0 coins and get a empty array in return. Less logic which means its easier to read. And yes I have a strangely high logaf on this.

Screaper91 (approved):
Makes sense to me to allow calling `flipCoin` with `0` and returning `[]` without modifying `PlayerComponent.coinFlips` if a card wants to flip a variable amount of coins.

PR #1320, opened 2025-03-02
Add Account Syncing To Title Screen

zunda-arrow:
This also fixes an issue where login failures were not handled at all! Yay!  I also discovered why toasts don't show in games while I was working on this. I changed toasts to show up everywhere since I think all the devs thought that was how toasts worked, so now they will show up in games too.
The visuals for this are NOT ready for users. This will need prettification from someone else.


zunda-arrow:
It seems there is a bug with toasts where the messages show up again when you switch screens. I think the component is being reloaded in places it shouldn't be.

zunda-arrow:
> It seems there is a bug with toasts where the messages show up again when you switch screens. I think the component is being reloaded in places it shouldn't be.

fixed

PR #1321, opened 2025-03-03
Optimize Root of App

zunda-arrow:
These are a few free optimizations. By moving where we grab these things from the redux store we will have to reload much less on each change.


alicetouhou:
LGTM

alicetouhou (approved):

PR #1325, opened 2025-03-03
Play screen fixes

Sense101:


alicetouhou:
I just made some changes to the play screen to fix some mobile issues, so now I'm wondering if your changes change how it looks on mobile? Mobile looked fine so it probably shouldn't

Sense101:
The only change to mobile in this pr is changing the spacer to max-height: 2vh;

alicetouhou (approved):
Just ran it to confirm all the spacing is still correct

PR #1328, opened 2025-03-03
Remove deck selected toast on return to main menu

Sense101:
No longer needed, this all works automatically now.


zunda-arrow (changes requested):
I'm not sure this is going to work. I just made some toast changes yesterday so now toasts do not reload when the page is changed in the router. The reason this change was made is to stop the entire virtual dom from rerendering when a toast is published.

zunda-arrow:
Nope i misunderstood what this does, lgtm

zunda-arrow (approved):

PR #1333, opened 2025-03-03
Primitive System For Banning Users

zunda-arrow:
This system is very primitive, and only bans the UUID. This means it can be easily bypassed.
## MIGRATION
```sql
ALTER TABLE users ADD COLUMN banned boolean DEFAULT false NOT NULL;
```


Tyrannicodin (approved):

alicetouhou (changes requested):
I think you "You are banned" message should not have a frowning face, also I think the readme description should probably be a bit more formal. You don't need to say it's unfortunate the system isn't good, just explain how it works

alicetouhou (approved):

PR #1336, opened 2025-03-04
Fix issues with new dropdown component

alicetouhou:
Turns out I forgot to remove an event listener, which it why it wasn't working before, oops.
The other change is to make the show-dropdown button clickable to hide the dropdown, as it's more intuitive.


zunda-arrow:
linted this for u

zunda-arrow (approved):

PR #1338, opened 2025-03-04
Achievements page

alicetouhou:
This PR finishes up the achievements page


zunda-arrow:
The only issue is "NO TITLE" has a tooltip when it should not

zunda-arrow:
I fixed the issue

PR #1339, opened 2025-03-04
Optimize Slot Empty Query

zunda-arrow:
This PR optimizes the `slot.empty` query to run in `O(1)` time. Now that we are getting more players, I am finally looking back to optimizing this system at the cost of complexity.
Note that now we can't link multiple cards to one slot, but we didn't do that a single time since the system was made so I seriously doubt this is important.

### DIFF
--- a/common/cards/advent-of-tcg/attach/brewing-stand.ts
+++ b/common/cards/advent-of-tcg/attach/brewing-stand.ts
@@ -49,7 +49,7 @@ const BrewingStand: Attach = {
 					if (!pickedSlot.inRow()) return
 
 					pickedSlot.row.heal(50)
-					pickedSlot.getCard()?.discard()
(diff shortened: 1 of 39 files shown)


Tyrannicodin (approved):
Code looks good, I trust you it makes it quicker.
wtf is a double question mark?

alicetouhou:
> Code looks good, I trust you it makes it quicker. wtf is a double question mark?

I think it's the same as || but it doesn't treat '' and 0 as false

Sense101:
I still have no idea how the component system works, so can't review this one

zunda-arrow:
@Screaper91 I would like you to be aware of this change. Also, if you are interested in further optimizing the performance of this system feel free to ping me on Discord.

Screaper91 (approved):

PR #1340, opened 2025-03-04
Stats page rework

alicetouhou:
- Fixes a lot of NaNs that display when you played 0 games
- re-orders statistics page to be easier to read, and to put the more important stats first


zunda-arrow (approved):
Tested and works!

PR #1343, opened 2025-03-04
Disable Start Game Buttons If Deck Is Not Selected

zunda-arrow:


zunda-arrow:
@Sense101 I think you would want to check this!

Sense101 (approved):
Looks good!

PR #1344, opened 2025-03-04
Low Hanging Fruit Game Performance Optimizations

zunda-arrow:
This PR reduces renrenders of the game board. It seemed to make modals open a little faster for me but i could be imagining that. It will definitely reduce power consumption though.

## Further Work
Improve the way status effects are read to reduce rerenders of the game board when a status effect is changed.

### DIFF
--- a/client/src/app/game/board/board-row.tsx
+++ b/client/src/app/game/board/board-row.tsx
@@ -50,9 +50,10 @@ function getRowState(playerEntity: PlayerEntity, rowEntity: RowEntity) {
 	}
 }
 
-function getActiveRow(playerEntity: PlayerEntity) {
+function isActiveRow(playerEntity: PlayerEntity, rowEntity: RowEntity) {
 	return (state: RootState) => {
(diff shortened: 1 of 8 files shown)


client/src/app/game/board/board-row.tsx
```diff
@@ -27,34 +32,54 @@ const getSlotByLocation = (
 
 type BoardRowProps = {
 	type: 'left' | 'right'
-	player?: PlayerEntity
+	boardForPlayerEntity: PlayerEntity
```
Tyrannicodin, line 35: This name is a bit confusing

Tyrannicodin (commented):
I can't actually confirm if this speeds up the site, but code looks functional.

client/src/app/game/board/board-row.tsx
zunda-arrow: I am going to disagree with you because "boardForPlayerEntity" is very specific, but if you elaborate on what you think is confusing I can adjust it.

alicetouhou (approved):

PR #1347, opened 2025-03-04
Achievement toasts in game

Tyrannicodin:


zunda-arrow:
I believe backgrounds don't show at the correct size in the tooltip but otherwise this should be good.

zunda-arrow:
I made a change so that we show the achievement icon in the toast when its needed.

zunda-arrow:
The acheivement icons in the toast are completely messed up.

PR #1360, opened 2025-03-06
Change direction of update modal

zunda-arrow:


Tyrannicodin (commented):
I don't really see the point of this

Sense101 (approved):
I think this is a good idea, scrolling upwards is confusing.

PR #1364, opened 2025-03-06
Clean up login animation

Sense101:
Changed it to just a fade when you switch to sync device and back. Looks cleaner.


alicetouhou:
I prefer the old animation but if you fix e2e I don't really care either way

zunda-arrow:
This login error does not show up on the sync accounts page. This needs to be fixed.
I will fix the e2e tests though.

zunda-arrow:
I get stuck after pressing "sync" with incorrect credentials. The retry button does not work either. I think this is why e2e tests fail.

PR #1365, opened 2025-03-06
Fix filter width being limited

Sense101:


alicetouhou:
Thanks, looks like this fixing the deck editor bug too

PR #1367, opened 2025-03-06
Add "Gotta Schreep" Achievement and Reward

zunda-arrow:
I added some tests for this so its gotta work


alicetouhou:
I feel like Gotta Schreep would be a better name

common/achievements/gotta-screep.ts
```diff
@@ -0,0 +1,39 @@
+import {RowComponent} from '../components'
+import query from '../components/query'
+import SleepingEffect from '../status-effects/sleeping'
+import {achievement} from './defaults'
+import {Achievement} from './types'
+
+const GottaSchreep: Achievement = {
+	...achievement,
+	numericId: 54,
+	id: 'gotta_screep',
```
Tyrannicodin, line 10: typo

common/achievements/gotta-screep.ts
```diff
@@ -0,0 +1,39 @@
...
+	levels: [
+		{
+			name: 'Gotta Screep',
```
Tyrannicodin, line 13: same as above

common/achievements/gotta-screep.ts
```diff
@@ -0,0 +1,39 @@
...
+			description: 'KO a Hermit while your active Hermit is sleeping.',
+			steps: 1,
+		},
+	],
+	onGameStart(game, player, component, observer) {
+		let isSleeping = false
+
+		observer.subscribe(player.hooks.onTurnStart, () => {
```
Tyrannicodin, line 21: I think this should just be checked on KO

Tyrannicodin (changes requested):

common/achievements/gotta-screep.ts
zunda-arrow: That does not work. If you KO on the same turn bdubs falls asleep, it actually counts as bdouble0 being asleep because the sleep status effect is applied before the KO is checked. The second test tests this case.

Tyrannicodin (approved):

PR #1370, opened 2025-03-06
Prevent spec and replay rematch

Tyrannicodin:


server/src/routines/matchmaking.ts
```diff
@@ -1186,6 +1199,8 @@ export function* createReplayGame(
 		gameEndTime: Date.now(),
 	})
 
+	broadcast([viewer.player], {type: serverMessages.SEND_REMATCH, rematch: null})
```
zunda-arrow, line 1202: Im confused, why does this need to be sent? I think that you shouldn't have the option to rematch if no rematch message is sent?

Tyrannicodin: I don't think it does, idk why I did this.

PR #1393, opened 2025-03-08
Moved checks around to stop Evil X crashing the server

NikoUY:


alicetouhou:
lgtm now that e2e passes! tysm

PR #1397, opened 2025-03-08
Fix toasts and don't show toasts for already complete achievements

Tyrannicodin:


zunda-arrow (approved):
Actually I probably don't need to run this, the code looks good

PR #1398, opened 2025-03-08
Better global achievement progress

Tyrannicodin:


zunda-arrow (approved):
I am assuming this is related to the bot API so I trust it works how you want it to.

PR #1401, opened 2025-03-09
Golden Apple to 2 Tokens

Rvtar:
**Description:**
Golden Apple in general is severely underused, and reducing its cost to 2 tokens might be able to offset the AFK-only drawback.

**Original Suggestor:**
Elementl0rd

**Supported By:**
mekktor, TheThunderW0lf, Aetherion, shortweekly, Rvtar, theking0fgg


zunda-arrow (commented):
Thanks for the patch and using the template! This can be merged once the tests pass. It is probably because a starter deck uses golden apple so it will need its token cost increased.

alicetouhou:
image
indeed is that

zunda-arrow (approved):

PR #1404, opened 2025-03-09
Fix EvilX Achievments

zunda-arrow:
Reworks evil x achievments to not rely on the `isEvilXGame` boolean. This boolean is set after achievements are added which causes issues. I think this system is better as well.

### DIFF
--- a/common/achievements/no-derpcoins.ts
+++ b/common/achievements/no-derpcoins.ts
@@ -16,7 +16,6 @@ const NoDerpcoins: Achievement = {
 	],
 	onGameStart(game, player, component, observer) {
 		const cost = getDeckCost(player.getDeck().map((card) => card.props))
-		console.log(cost)
 		if (cost > 0) return
 		observer.subscribe(game.hooks.onGameEnd, (outcome) => {
 			if (outcome.type !== 'player-won') return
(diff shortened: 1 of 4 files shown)


common/achievements/no-derpcoins.ts
```diff
@@ -17,12 +15,8 @@ const NoDerpcoins: Achievement = {
 		},
 	],
 	onGameStart(game, player, component, observer) {
-		if (!game.state.isEvilXBossGame) return
-		const cost = getDeckCost(
-			game.components
-				.filter(CardComponent, query.card.player(player.entity))
-				.map((card) => card.props),
-		)
+		const cost = getDeckCost(player.getDeck().map((card) => card.props))
```
Tyrannicodin, line 18: I think @Screaper91 used that component query for a reason

common/achievements/no-derpcoins.ts
```diff
@@ -17,12 +15,8 @@ const NoDerpcoins: Achievement = {
...
+		console.log(cost)
```
Tyrannicodin, line 19: Excess log

Tyrannicodin (changes requested):

common/achievements/no-derpcoins.ts
zunda-arrow: This should be the same thing. I think `getDeck()` is easier to look at.

Tyrannicodin (approved):

PR #1408, opened 2025-03-09
Fix "Channeling" achievement

Screaper91:
- Fixes achievement only granting if AFK Hermit is knocked-out and _base_ damage is strictly greater than active row health
- New description: "While your active Hermit has red HP, use Lightning Rod to redirect an attack's damage away from them."

### DIFF
--- a/common/achievements/channeling.ts
+++ b/common/achievements/channeling.ts
@@ -1,6 +1,6 @@
 import LightningRod from '../cards/attach/lightning-rod'
 import {CardComponent} from '../components'
-import {afterAttack} from '../types/priorities'
+import {afterAttack, beforeAttack} from '../types/priorities'
 import {achievement} from './defaults'
 import {Achievement} from './types'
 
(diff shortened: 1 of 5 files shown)


Screaper91:
I am not going to adjust this to consider Trapdoor, rare Farmer Beef, or rare Overseer because those interactions would add even more complexity and may change in the future.

alicetouhou:
If you think this is good, I approve in that case.
Really surprised at how complicated this achievement ended up being, thank you for fixing it

alicetouhou (approved):

zunda-arrow:
Although this implementation works, I dont think its wrong to say its an unwieldy behemoth. I feel this achievement is far too reliant on the current code of the game for me to feel like we should take the maintnence burden of this. I think I was wrong to add the achievement with this description initially. The options I feel are more reasonable are:
- Redo attacks so we can grab more info
- Change the achievement to "While your active hermit has Red HP, use lightning rod to redirect an attack away from them"

zunda-arrow (approved):
Thanks for updating the requirements! I'm really happy you put the work in to implement the original version of this achievement, because I think we would've have known how unwieldy the code would be otherwise.

PR #1414, opened 2025-03-09
Etho UR Costs 2 Tokens With Only PvP/Wild Items

Rvtar:
**Description:**
Etho UR's usefulness mainly comes from the fact that he is compatible with items of any type, allowing him to be used in almost any deck, especially multitype decks, thus his cost of 3 tokens. However, this makes the card a hindrance to include in decks of his type, PvP due to his high cost with practically zero value out of his universal compatibility. This conditional cost reduction will allow him to be more useful in PvP decks, which are currently underpowered in the current meta, while remaining balanced in the general context.

**Original Suggestor:**
Elementl0rd (Token Cost), theking0fgg (PvP Clause)

**Supported By:**
eights8s, TheThunderW0lf, Aetherion, Rvtar, shortweekly, Jacksimpson2077


Tyrannicodin (commented):
Looks good from a code standpoint, I'm not sure on adding another exception to tokens.

common/types/cards.ts
```diff
@@ -1,7 +1,7 @@
 import {GameModel} from '../models/game-model'
 
 export type CardRarityT = 'common' | 'rare' | 'ultra_rare'
-export type TokenCostT = -1 | 0 | 1 | 2 | 3 | 4 | 5 | 'wild'
+export type TokenCostT = -1 | 0 | 1 | 2 | 3 | 4 | 5 | 'wild' | 'etho'
```
zunda-arrow, line 4: It would be better to call this `etho-ur`

zunda-arrow (changes requested):
Looks good

alicetouhou:
> Looks good from a code standpoint, I'm not sure on adding another exception to tokens.

I understand where you are coming from but I don't agree it's a good idea to block changes because we think they might be bad, after the most active community members have agreed they are ok.

I would suggest implementing this if there's 6 or more supporters (discluding me), and if the confusion issues do arise, we reduce it. Just need to wait until rvtar modifies the description

zunda-arrow (approved):

Rvtar:
> Just need to wait until rvtar modifies the description

I already did!

Rvtar:
Done. (hard facepalm)

zunda-arrow (approved):

PR #1416, opened 2025-03-09
Fix Wipeout Achievement

zunda-arrow:
Fixes the wipeout achievement not counting when the game ends after knocking out all the opponent's hermits.


alicetouhou:
I think there needs to be a "killed by lava bucket" test.
I would also update the description to `Knockout 3 Hermits in the same turn. Includes both your turn and your opponent's following turn.`

PR #1427, opened 2025-03-10
Optimize Queries For Hermit In Rows

zunda-arrow:
This breaks advent impulse which I will need to figure out


Screaper91 (approved):
I fixed the remaining issues, let me know if `getSupportingItems` needs changes

PR #1431, opened 2025-03-10
Make Efficiency Actually Last The Whole Turn

Rvtar:
The description says "for this turn". Therefore, one should be able to attack multiple times without necessary items if they have used efficiency this turn i.e. Efficiency should not remove itself after an attack.

### DIFF
--- a/common/status-effects/efficiency.ts
+++ b/common/status-effects/efficiency.ts
@@ -16,15 +16,6 @@ const EfficiencyEffect: StatusEffect<PlayerComponent> = {
 		})
 
 		observer.subscribeWithPriority(
-			game.hooks.afterAttack,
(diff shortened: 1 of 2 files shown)


Tyrannicodin (approved):

zunda-arrow:
we need @Screaper91's review on this I think

common/status-effects/efficiency.ts
Screaper91, line 1: Please add the following to prevent rare Gem stacking copies of this status
```ts
	applyCondition: (_game, value) =>
		value instanceof PlayerComponent &&
		!value.hasStatusEffect(EfficiencyEffect),
```

zunda-arrow:
Ill finish this PR

Screaper91 (approved):

PR #1432, opened 2025-03-10
Array of Structs Optimization

zunda-arrow:
Split different component types into their own table to reduce how much needs to be searched per query


zunda-arrow:
@Screaper91 i would appreciate if you could review this

Screaper91 (approved):

PR #1436, opened 2025-03-11
Add Tag Team Achievement

zunda-arrow:


common/achievements/tag-team.ts
```diff
@@ -0,0 +1,30 @@
...
+import {Achievement} from './types'
+
+const TagTeam: Achievement = {
+	...achievement,
+	numericId: 55,
+	id: 'tag_team',
+	levels: [
+		{
+			name: 'Tag Team',
+			description: 'Win a game after swapping your active Hermit 6 times.',
```
Tyrannicodin, line 11: This is ambiguous, is it in one game or across 6 games? Do you have to win on the same turn as when you swap?

PR #1438, opened 2025-03-11
Fix misusage of `PlayerComponent.getDeck`

Screaper91:
`getDeck` returns cards in a player's draw pile, not their full deck


zunda-arrow (commented):
THANK YOU! I did not realize this. Can you rename the function to `getDrawPile` as well?

PR #1445, opened 2025-03-19
Make Cards in Frozen Hand Slots Unplayable

Rvtar:
Literally. There might be some usefulness in making cards like for Advent.

### DIFF
--- a/server/src/routines/turn-actions.ts
+++ b/server/src/routines/turn-actions.ts
@@ -162,6 +162,11 @@ export function playCardAction(
 		'You can not play a card in a slot with a card in it',
 	)
 
+	assert(
+		!game.components.filter(SlotComponent, query.slot.frozen).includes(pickedSlot),
+		'You cannot interact with cards in frozen slots'
+	)
+
 	const row = pickedSlot.row
 	const rowIndex = pickedSlot.index
 	const player = pickedSlot.player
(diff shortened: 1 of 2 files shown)


zunda-arrow:
Looks good! I would like @Screaper91's review and a test.

Rvtar:
How do I just freeze a hand slot?

alicetouhou:
> How do I just freeze a hand slot without a card?

Shouldn't be possible because hand slots don't exist until the card is added to them. I assume you are trying to make them frozen as soon as the slot is created? there's probably a hook you can use for that

Screaper91 (changes requested):
This works for the current client, however this doesn't invalidate using those cards if the server receives an play card action (modified clients/testing). Add an assertion to `server/src/routines/turn-actions.ts > playCardAction` that `card.slot` is not frozen.

zunda-arrow:
Nice catch!

server/src/routines/turn-actions.ts
```diff
@@ -162,6 +162,11 @@ export function playCardAction(
 		'You can not play a card in a slot with a card in it',
 	)
 
+	assert(
+		!game.components.filter(SlotComponent, query.slot.frozen).includes(pickedSlot),
+		'You cannot interact with cards in frozen slots'
```
Screaper91, line 167:
```suggestion
		!query.slot.frozen(game, card.slot),
		'You cannot play cards that are in frozen slots'
```
Is this your intention?

Rvtar: I think so.

Screaper91 (approved):

PR #1447, opened 2025-03-20
Fix Poe-Poe Enforcer Incrementing Multiple Times from Extra Attacks Hitting a 0 HP Hermit

Rvtar:
Also found the test for this one failing in Scuffed branch.


Rvtar:
I screwed the history so hard.

PR #1448, opened 2025-03-20
Fix Poe-Poe Enforcer Incrementing Multiple Times from Extra Attacks Hitting a 0 HP Hermit

Rvtar:
Also found the test for this one failing in Scuffed branch.


Screaper91 (changes requested):
`deadTargets: Array<RowComponent>` should be reset occasionally to handle using Curse of Binding to knock out the same row multiple times in the same game.

Screaper91 (approved):

PR #1451, opened 2025-03-22
Only Allow Primary Attack With Target Block

zunda-arrow:
I do not know who suggested this first.

The goal of this change is to reduce the strength of target block to something more manageable.


Tyrannicodin:
Closed in favour of a one turn delay

PR #1452, opened 2025-03-25
Update impulsesv-rare 2. Attack.ts

AGentW29:
Added missing check to see if BoomerBdubs or SteampunkTango are on the board to increase Impuls secondary attack


alicetouhou:
Appreciate the contribution but this is not a bug!
We do want to make the description more obvious that it does not include AE cards, though

AGentW29:
Ok. Thank you for replying. I will see if I can come up with an elegant wording to make the description more clear. Do I have to fill out a pull request template for that or is that only for balance patches?

alicetouhou:
You only need to fill out the template if you're changing a card's behavior.

AGentW29:
ok thank you

PR #1453, opened 2025-03-25
Clarification of 2. Attack.ts

AGentW29:
Added "non alter ego" to 2. Attack description for clarification


zunda-arrow (changes requested):
I think the wording "base set" would be better because we use this for Jack of All Cards. I think it would awesome if you made this change for NHO too!

PR #1457, opened 2025-03-25
Make only new achievements sent over

alicetouhou:
This change makes it so that only achievements that have been changed since last login are sent to the client. This should help reduce server costs as decreases outgoing bandwidth


zunda-arrow:
Instead of sending over everything, maybe we should just send a hash. Obviously it won't reduce bandwidth in many cases but the games where achievement progress doesn't change it will help. This also won't cause any slowdowns on poor internet.

PR #1458, opened 2025-03-26
Fixes double counting.ts

AGentW29:
Added DISTINCT to ensure each card would only be counted once per deck (even if multiple cards exists) to prevent percentages over 100% (and to make them accurate)


zunda-arrow:
@alicetouhou will need to review this

alicetouhou:
thanks, looks good

AGentW29:
Ok. Do I automatically get notified, if my changes are accepted or declined, because I might have deleted my previous changes too fast. At least I don't see them in game... Or does it take a while until they are applied?

alicetouhou:
You'll need to wait until we make a new release to see them in game

PR #1462, opened 2025-03-26
Fix Evil X being unable to disable Puppetry if attacker has 0 afk Hermits

Screaper91:
Adds a `mode` parameter to `GameModel.addCopyAttackModalRequest` to allow different setup behavior for `EvilXisumaRare` and `ZombieCleoRare`/`RendogRare`

### DIFF
--- a/common/cards/hermits/evilxisuma_rare.ts
+++ b/common/cards/hermits/evilxisuma_rare.ts
@@ -70,7 +70,7 @@ const EvilXisumaRare: Hermit = {
 				)
 				if (!opponentActiveHermit?.isAlive()) return
 
-				game.addCopyAttackModalRequest(
+				game.addPickAttackModalRequest(
 					{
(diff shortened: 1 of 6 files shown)


zunda-arrow (commented):
I think that a better solution would be renaming the function to "pickAttackModal" and then have a boolean to specify if blocked actions should be included. Its the same thing but I think the functions purpose will be clearer.

Screaper91:
> I think that a better solution would be renaming the function to "pickAttackModal" and then have a boolean to specify if blocked actions should be included.

I prefer `mode: 'copy' | 'disable'` as "passive attacks" cannot be disabled and this is more easily extended for new cards that may need to select attacks.

zunda-arrow:
I'm pretty busy with school right now so I can't check this! Feel free to merge when you are happy with it, I trust your judgement.

PR #1464, opened 2025-03-27
Optimize Message Names in Production

zunda-arrow:


zunda-arrow:
This one can break prod if we aren't careful so I want to wait on this.

PR #1466, opened 2025-03-27
Power bed

Rvtar:


alicetouhou:
small change, "this card" instead of it, also no space before it

common/cards/attach/power-bed.ts
Tyrannicodin, line 1: Looks like you forgot to remove this file

common/cards/bed-update/attach/power-bed.ts
Tyrannicodin, line 1: This didn't appear to take health from the hermit.

Tyrannicodin (changes requested):

Tyrannicodin (approved):

PR #1470, opened 2025-03-27
Start Converting Logs To Quickwit

zunda-arrow:


zunda-arrow:
Although we switched off fly.io, this is still something I want to investigate

PR #1473, opened 2025-03-29
Load of beds

Tyrannicodin:
Add several beds


common/components/row-component.ts
```diff
@@ -100,9 +101,26 @@ export class RowComponent {
...
+			StatusEffectComponent,
+			query.effect.targetIsCardAnd(query.card.entity(hermit.entity)),
+		)
 		this.health = Math.min(
 			this.health + amount,
 			Math.max(this.health, hermit.props.health),
 		)
+		if (overhealed)
+			this.health = this.health + amount
+	}
```
Rvtar, line 114: This would make it heal twice, once with the cap, and if "overhealed", heal again without cap.

common/status-effects/lit_fuse.ts
```diff
@@ -0,0 +1,56 @@
...
+				let damage = 0
+				target.player.getHand().forEach((card) => {
+					if (![TNT.id, MinecartWithTNT.id].includes(card.props.id)) return
+					card.discard()
+					damage += 60
+				})
+				if (!target.slot.inRow()) return
+				const hermit = target.slot.row.getHermit()
+				if (damage === 0 || !hermit) {
+					target.discard()
```
Rvtar, line 42: Do you mean effect.remove()?

common/status-effects/lit_fuse.ts
```diff
@@ -0,0 +1,56 @@
...
+					return
+				}
+				target.slot.row.damage(damage)
+				game.battleLog.addEntry(
+					player.entity,
+					`$p${hermit.props.name}$'s $e${target.props.name}$ detonated, dealing $g${damage}hp$ damage.`,
+				)
+				target.discard()
```
Rvtar, line 50: Do you mean effect.remove()?

Rvtar (changes requested):

common/cards/bed-update/single-use/uncomfy-bed.ts
```diff
@@ -0,0 +1,47 @@
...
+	onAttach(game, component, observer) {
+		const {player} = component
+
+		observer.subscribeWithPriority(
+			player.hooks.afterApply,
+			afterApply.CLEAR_STATUS_EFFECT,
+			() => {
+				let removeFrom = game.components.filter(
+					StatusEffectComponent,
+					query.effect.is(SleepingEffect),
```
Rvtar, line 28: I made another one, PermanentSleepingEffect, for this update.

common/status-effects/lit_fuse.ts
Tyrannicodin: No, the bed should be removed when the fuse detonates, but I do also need to do effect.remove

Rvtar (approved):
LGTM

PR #1479, opened 2025-03-31
April fools 2025

Tyrannicodin:


common/utils/ranks.ts
```diff
@@ -7,6 +7,7 @@ export function getCardVisualTokenCost(
 	if (tokens === 'wild') return 1
 	if (tokens === 'etho-ur') return 3
 	if (tokens === -1) return 0
+	if (tokens === -0.5) return 0
```
alicetouhou, line 10: I would change these 2 if statements to `else return 0`

common/cards/bed-update/single-use/disappointing-bed.ts
alicetouhou, line 1: I would prefer if this card was changed to be balanced at 0 tokens

common/components/row-component.ts
```diff
@@ -100,9 +101,28 @@ export class RowComponent {
...
 		if (this.health === null) return
 		if (!hermit?.isHealth()) return
+		const overhealed = this.game.components.exists(
+			StatusEffectComponent,
+			query.effect.targetIsCardAnd(query.card.entity(hermit.entity)),
+		)
+		if (overhealed) {
+			this.health = this.health + amount
+			return
+		}
```
alicetouhou, line 111: What is the purpose of this change? I don't see why this isn't within the `overhealed` status effect file

alicetouhou (commented):
There seems to be some weird git issue bc msgpack is attributed to this PR. Wonder why

More importantly, I think automated testing should be added to at least the cards that are perceived to be more common, just so we can try to cut down on bugs.

Tyrannicodin:
> There seems to be some weird git issue bc msgpack is attributed to this PR. Wonder why
I have no idea

> More importantly, I think automated testing should be added to at least the cards that are perceived to be more common, just so we can try to cut down on bugs.
I do not have the energy to do this tbh (I've been ill today), I will just be getting the test (which I am struggling to figure out) working and leaving it to others.

common/components/row-component.ts
Tyrannicodin: It allows us to continue using the .heal() method, it would be difficult to do as we'd have to calculate the already given hp, and then add on the extra.

common/cards/bed-update/single-use/disappointing-bed.ts
Tyrannicodin: @Rvtar

Rvtar:
My original idea was to have Power Bed's penalty bypass damage reduction.

zunda-arrow:
I'm not sure about merging a few things in this PR and I would like screaper's reviewer before merging this, I think it would be better to build the docker image off of this branch instead of merging.

Tyrannicodin:
> I'm not sure about merging a few things in this PR and I would like screaper's reviewer before merging this, I think it would be better to build the docker image off of this branch instead of merging.

I don't know how to do this, but if you want to please do!

PR #1484, opened 2025-04-05
Fixing April Fools Fails

Rvtar:


Rvtar:
Idk why it's not running tests. I guess I should use a new branch instead.

PR #1514, opened 2025-04-21
New Boss AI

AGentW29:
Player like AI Boss. Single use cards don't work and need fixing, for both player and AI.


alicetouhou:
this is awesome!
if you need help i can help out

common/cards/boss/hermits/new_boss.ts
Screaper91, line 1: This is unnecessary and all references to `NewBoss: Hermit` should be removed unless there is new boss behavior for a special card designed.

common/utils/state-gen.ts
```diff
@@ -157,6 +157,12 @@ function setupEcsForPlayer(
 	cards.slice(0, amountOfStartingCards).forEach((card) => {
 		card.attach(components.new(HandSlotComponent, playerEntity))
 	})
+
+	// Create single-use slot with proper initialization
+	components.new(BoardSlotComponent, {
+		type: 'single_use',
+		player: playerEntity
+	}, 0, null)
```
Screaper91, line 165: This adds two additional single use slots to the game when only one should exist (see line 67)
```suggestion

```

server/src/routines/turn-actions.ts
Screaper91, line 1: Please revert all changes to this file! You removed checking the played card's attach condition, broke all single-use card functionality, and removing the single use card moves it to discard instead of hand.

server/src/routines/virtual/new-boss-ai.ts
Screaper91, line 1: Please look at [tests/fuzz/fuzz-ai.ts](url) for how to determine what actions are available for AI players and how they can be completed. A good starting point would be replacing all `choose(array, rng)` functions with a custom sorter to prioritize certain slots over others like `array.sort(prioritizeActiveRow).at(0)`.

Your implementation would not apply single use cards because `END_TURN` is always first in `availableActions` (before `APPLY_EFFECT`)

Screaper91 (changes requested):
Here are my comments on the changes to the backend to run this new AI. I'll let @zunda-arrow determine if the other changes to slot components and their queries are appropriate.

server/src/routines/virtual/new-boss-ai.ts
AGentW29: Thank you for the feedback. I will take a look at this as soon as I get a chance

server/src/routines/turn-actions.ts
AGentW29: I reverted the changes to this file and I adjusted common/utils/state-gen.ts, but now when I try to play single use cards, as soon as I place them they pop right back into my hand and I can't play them. That's what I was trying to fix in turn-actions.ts. Do you know what else could have caused the single use cards to stop working?

AGentW29: Nevermind. I found the error. Thanks for reviewing and helping me though.

AGentW29:
I finished the new boss ai now. I tested it a lot and I haven't come across any major bugs, but there might still be a scenario, where it doesn't work. The new boss ai uses different decks each time which I have sorted and tested. There is a description at the end on what cards to avoid when creating additional decks, but I probably didn't flag all game crashing cards. I am going to switch it to a normal pull request now, since it works pretty well now. If anyone sees any ways to improve what I have done feel free to modify anything. Btw. I don't understand what it means, that the CI / Lint (pull request) is failing.

Screaper91 (approved):
This looks good enough to continue work on the new branch so we can collaborate better.

alicetouhou:
> This looks good enough to continue work on the new branch so we can collaborate better.

sounds good!

PR #1515, opened 2025-04-29
Improved AI deck and fixed compiling errors

AGentW29:
Switch TangoTekRare to SteampunkTangoRare to match Terraform deck. Fixed 2 small compiling errors.


server/src/routines/matchmaking.ts
```diff
@@ -214,7 +214,7 @@ function* gameManager(con: GameController) {
 			)
 
 			achievements.forEach((achievement) => {
-				achievement.props.onGameEnd(con.game, player, achievement, outcome)
+				achievement.props.onGameEnd(con.game, player!, achievement, outcome)
```
Screaper91, line 217: This should not use `player!` to assert the player component. If your environment doesn't recognize that `player` is not `null`, try switching `let player = ` for `const player = `

server/src/routines/virtual/new-boss-ai.ts
```diff
@@ -987,7 +987,7 @@ function getNextTurnAction(
 		return [
 			{
 				type: selectedAttackType,
-			},
+			} as AttackActionData,
```
Screaper91, line 993:
```suggestion
			} satisfies AttackActionData,
```
`as` overrides the type regardless of value, `satisfies` asserts the value does match the provided type

Screaper91 (changes requested):
Please do not work off of `hc-tcg:master`, make future contributions branched from `hc-tcg:ai-development` to reduce merge conflicts.

Screaper91 (approved):
Please fix

zunda-arrow (approved):
LGTM, not exactly sure what this was for but I trust screaper lol

PR #1516, opened 2025-05-15
Remove Redux Saga From Game Code

zunda-arrow:
Redux saga is a mess to work with and it has worked its way into our tests. I think its very slow and it making the tests run very slow. Second, this makes the game code easier to compose with other parts of the project.

Fuzz testing:
Removing redux saga from fuzz tests reduced their time by about a second, maybe thats around 15%.


zunda-arrow:
This works and I think is a good idea, but I will need to redo how tests work

PR #1517, opened 2025-05-23
Possible fix for connection timeouts

zunda-arrow:
The theory is that something is going wrong with the fly proxy and clients are losing connection, but they do not know that they are losing connection and are never connected. This forces the clients in the pool to be refreshed every two minutes.


zunda-arrow:
We don't need this anymore. The new servers shouldn't have this issue.

PR #1518, opened 2025-05-28
Fix Diamond Armor Blocking Effects Separately

Rvtar:


Screaper91 (approved):
`tests\unit\game\effects\diamond-armor.test.ts` should be updated to test an affected scenario

Rvtar:
Add the test with double SU version since it's much simpler.

PR #1522, opened 2025-06-05
Fix achievement progress not updating database

Screaper91:
The database now updates achievement progress with all goal updates from a game instead of only the last goal update.
- Fixes "How Did We Get Here" not being awarded if the game continues after having at least 5 statuses
- Fixes "SUStainable" only counting up to one Composter used per game


zunda-arrow (approved):
LGTM!

PR #1529, opened 2025-07-02
Add Comments for a Smarter AI.

Rvtar:
My comments have prefixes of SMART:
There's one question I labeled with Q:
I hope I didn't miss anything.

### DIFF
--- a/server/src/routines/virtual/new-boss-ai.ts
+++ b/server/src/routines/virtual/new-boss-ai.ts
@@ -75,6 +75,7 @@ function getNextTurnAction(
 
 	// Check if we have no active hermit but have AFK hermits
 	// This could happen if our active hermit was just knocked out
+	// SMART: Switch to most turns alive vs. current hermit (/+damage SU conditional) / quickest KO, has enough energy to attack / [variants] key hermits.
 	if (!activeHermit && hermitsOnBoard.length > 0) {
 		console.log(
 			'New Boss AI - No active hermit found, switching to hermit with most health',
@@ -160,6 +161,7 @@ function getNextTurnAction(
 		if (!pickRequest) return []
 
 		// Find a valid slot to pick based on the pick request's canPick query
+		// SMART: Check what request it is and respond accordingly.
 		const validSlot = game.components.find(SlotComponent, pickRequest.canPick)
 
 		if (validSlot) {
@@ -204,6 +206,7 @@ function getNextTurnAction(
 		if (modalRequest.modal.type === 'copyAttack') {
 			console.log('New Boss AI - Handling copyAttack modal')
 			// Randomly choose between primary and secondary attack
+			// SMART: Evaluate positions.
 			const availableAttacks = modalRequest.modal.availableAttacks
 			if (availableAttacks.length > 0) {
 				const randomIndex = Math.floor(Math.random() * availableAttacks.length)
@@ -232,6 +235,7 @@ function getNextTurnAction(
 			const selectionSize = modalRequest.modal.selectionSize
 
 			// If no selection needed, just confirm
+			// SMART: At least make it handle spyglass(simplest: token cost), chest, HHH, and say yes for Grian - Borrow.
 			if (selectionSize === 0) {
 				console.log('New Boss AI - No selection needed, confirming')
 				return [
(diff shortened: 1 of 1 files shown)


server/src/routines/virtual/new-boss-ai.ts
```diff
@@ -295,6 +299,7 @@ function getNextTurnAction(
 			console.log('New Boss AI - Handling dragCards modal')
 			// For drag cards, we'll just return the cards as they are
 			// This is a simple approach that should work for most cases
+			// SMART: I think there's just Brush rn and it's gonna be a lot of work for one specific case.
```
Tyrannicodin, line 302: Considering brush isn't even in base cards, probably safe to ignore.

server/src/routines/virtual/new-boss-ai.ts
```diff
@@ -334,6 +340,8 @@ function getNextTurnAction(
 			console.log('New Boss AI - Current active hermit health:', activeHealth)
 
 			// If active hermit has less than 90 health, try to switch to a healthier one
+			// SMART: DON'T, THERE ARE ONLY SELECT CASES WHERE YOU WANT TO DO THIS OUTSIDE OF LAST LIFE.
+			// Q: Cant' we combine the functions with the previous one?
```
Tyrannicodin, line 344: Not sure what the previous one is, if you mean the check for no active hermits at the start of the action, then yes it would be a good idea.

server/src/routines/virtual/new-boss-ai.ts
```diff
@@ -398,6 +406,7 @@ function getNextTurnAction(
 
 		if (hermitCards.length > 0) {
 			// Find all empty hermit slots where we can place hermits
+			// SMART: Look for AFK damage possibilities (i.e. Hotguy's on the oppoenet's board and it's not in our deck / Bow or corssbow played before.)
```
Tyrannicodin, line 409: This would be tricky as we'd need a log of all cards played, maybe we allow the ai to see the opponent discard pile?

server/src/routines/virtual/new-boss-ai.ts
```diff
@@ -1118,6 +1137,7 @@ function getNextTurnAction(
 			// Handle special cases for different action types
 			if (randomAction === 'CHANGE_ACTIVE_HERMIT') {
 				// Find a random hermit to switch to
+				// SMART: Anything but random.
```
Tyrannicodin, line 1140: I'd suggest most health, token cost, item power, maybe with some sort of formula to evaluate?

server/src/routines/virtual/new-boss-ai.ts
```diff
@@ -1189,6 +1209,7 @@ function getNextTurnAction(
 					return []
 				}
 			}
+			// SMART: YOU CAN ALSO PLAY A SU WITH GEM
```
Tyrannicodin, line 1212: Isn't this just a normal play SU action?

Tyrannicodin (commented):
Reminder that this AI is supposed to be pharaoh so it could possibly be tuned to play better with the pharaoh deck.

server/src/routines/virtual/new-boss-ai.ts
Rvtar: Yeah that's the one I meant.

server/src/routines/virtual/new-boss-ai.ts
Rvtar: I'm thinking of doing priorities first then maybe probability-related scoring.

zunda-arrow: I thought about this in the past and this was my idea:
1. If my hermit is Pharaoh and I have more than one life, do nothing.
2. If my hermit is under, N HP, and I have a Pharaoh on the board, switch to Pharaoh.
3. If my hermit is under N HP and I have one life, switch to my strongest hermit (strongest can be most token cost)
4. Otherwise, do nothong

zunda-arrow:
This feels to general to me. We should only handle what we are planning to put in Pharaohs deck. Then, we can handle other cards if there is another boss in the future.

zunda-arrow (approved):

AGentW29:
Originally I did want to create a pharaoh deck, but I then switched to create a general AI player, because I thought it would be more fun to play against

server/src/routines/virtual/new-boss-ai.ts
AGentW29: I didn't add any bow's or crossbows to the decks, because I couldn't get it to work in a sensible way. Same with the spyglass. Since the AI doesn't get any information from using the spyglass I just left it out of the decks.

PR #1530, opened 2025-07-04
Elo UI

Rvtar:
Displays for elo and a few more things for it to function. The mechanic itself is not in here.


alicetouhou:
this looks good! I will make a PR to your branch when it's ready. of course I'm still deciding exactly how I want the system to work, so if you have feedback or whatever please tell me

PR #1535, opened 2025-07-08
Optimize AI

Screaper91:
Reduces size of `new-boss-ai.ts` with minimal changes to action behavior and initial preparations for Pharaoh boss use.
- Reduces AI attempting invalid actions
- Removes timeout logic for returning `[]`
- Promotes using queries and/or component members


Screaper91:
I converted `getNextTurnAction`'s return type from an array to a single action for safer code, but there may be special cases with pick/modal requests that may be better handled at once (e.g. Piston, Chest)

zunda-arrow (approved):
LGTM

PR #1536, opened 2025-07-15
Changes To Make Game Serializable

zunda-arrow:
The idea is to not use captured variables in functions because we can't really serialize that.
This creates a whole in the typing for the game but I think it won't be a big deal.

I want to make this change because it will be easy to move what server a game is stored on in the middle of a game to keep it close to the active player for example, or maybe not that crazy but that should get the idea across.

We do also have the `InstancedValue` class. Should we use that instead? Im not convinced by it though.


Rvtar:
I think Weakness currently has something similar. You should probably look into effects too.

zunda-arrow:
I really like this approach but I don't think its possible right now.

PR #1537, opened 2025-07-17
Mumbo Base Set

Rvtar:
Midsummer 2025 Patch #1: Mumbo's double damage condition only triggers with base set pranksters and change to 3 tokens.

Suggestor: Aetherion
No Opposition: Jacksimpson2077, Rvtar, shortweekly, Benji


alicetouhou:
personally, I would prefer the qualification to be rare instead of base set

Rvtar:
I think there have been complaints about Mumbo with Keralis, so probably not.

alicetouhou:
This is a small detail, but would you be ok with changing `Prankster from the Base Set expansion` to `Base Set Prankster`? It's just less words.

PR #1538, opened 2025-07-18
Impulse Requires Both

Rvtar:
Midsummer 2025 Patch 02: Impulse now requires both Tango and Bdubs for +80 damage. Cost reduced to 3.

Suggestor: Jacksimpson2077
No Opposition: Rvtar, shortweekly, Aetherion, Benji


alicetouhou:
I'm still not sure how I feel about this one. I'd like other people to give input. It seems split

PR #1546, opened 2025-07-21
Event propagation netcode

zunda-arrow:
This PR switches the game to use an event propagation system. This means that each turn action is broadcast to the server then each player. This allows clients to run the turn action before asking the server what it does.

The PR hides certain information from the clients. The opponent's hand and deck are hidden.


zunda-arrow:
I am up to fixing type issues and working on getting all tests to pass. After that, we will need a test server because I expect there will be a lot of broken things.

PR #1548, opened 2025-09-04
Add 1 turn delay to target block

Tyrannicodin:


Tyrannicodin:
Closed, #1540 already exists I just missed it.

PR #1549, opened 2025-09-04
Fast forward meta-update-july-2025

Tyrannicodin:
I'm pretty sure this is ok to merge straight away but I'll wait until the morning to double check (or someone can merge it earlier)


Tyrannicodin:
Can't normal merge, so not going to go for it. Will rebase my other pr.

PR #1560, opened 2025-09-21
Add secret resetting

Tyrannicodin:
Adds a button with a confirm modal to reset the user secret. This logs the player out of all devices except the one they are currently on.

This also fixes a bug where syncing whilst logged in would not immediately log you back in (you had to refresh)

Re-opened due to me nuking my fork

### DIFF
--- a/client/src/app/main-menu/settings.tsx
+++ b/client/src/app/main-menu/settings.tsx
@@ -478,7 +478,7 @@ function Settings({setMenuSection}: Props) {
 													onClose={closeModal}
 												>
 													<p className={css.warning}>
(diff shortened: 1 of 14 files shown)


client/src/app/main-menu/settings.tsx
```diff
@@ -252,7 +256,7 @@ function Settings({setMenuSection}: Props) {
 							variant="default"
 							onClick={() => setModal(null)}
 						>
-							Nevermind
+							Never mind
```
zunda-arrow, line 259: Eh?

PR #1565, opened 2025-09-27
Space Race achievement

Tyrannicodin:


Screaper91 (commented):
Test achievement progress when Renbob has type advantage (probable double counting) and that the `Record` is partially cleared when Renbob leaves the board as if it is not "the same Renbob Rare card".

zunda-arrow (changes requested):
Just needs tests! I will write them.

PR #1569, opened 2025-09-27
Turtle Master achievement

Screaper91:
Includes reward and tests for exact behavior


Tyrannicodin:
I like the name, for reward I'm thinking blue glazed terracotta hearts

PR #1576, opened 2025-10-07
Don't reload infinitely on invalid secret or user id

zunda-arrow:
This should make the game easier to develop. Should also fix issues if someone messes with their local storage.


Tyrannicodin (approved):
@Screaper91 Unsure if this should be merged, can you check please?

